PR #118, opened 2015-05-17
fix(make): fix clean target for docs/cli

worr:
clean-all-docs and clean-all-cli aren't valid targets. The current mako
template causes `make clean` to abend reporting that it can't make these
targets.


Byron:
Thank you ! Please note that I have fixed the issue a little differently, nonetheless, it should be working now.

PR #150, opened 2016-04-25
fix(dependencies): Pin url dependency to "= 0.5"

dermesser:
The version bump of url to 1.0 broke URL encoding serialization (see
also
https://github.com/Byron/yup-oauth2/pull/24#issuecomment-214455146)

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -6,7 +6,7 @@ DO NOT EDIT !
 The MIT License (MIT)
 =====================
 
-Copyright © `2015` `Sebastian Thiel`
+Copyright © `2015-2016` `Sebastian Thiel`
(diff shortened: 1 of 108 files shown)


Byron:
Thanks for this PR ! I have made the change in `etc/type-api.yaml` and will push it soon. This will also mean the code-change presented here will be replaced eventually.

It seems like the change was done on all files manually, and I recommend only editing the `*.mako` files instead. Something like `make regen-apis` would then rebuild all the code accordingly.

dermesser:
thanks! I did it almost the way you're suggesting -- I edited etc/api/type-api.yaml. For some reason, the url dependency is defined there.

PR #154, opened 2016-09-21
URL-encoding '/' in URLs is not accepted by Google APIs

dermesser:
I'm not sure if this is the best solution, or valid for every API, I haven't dug too much in the code generation templates (you probably know best!)

I hit this issue when writing an example client for Google Cloud Pubsub: Hitting the URL `https://pubsub.googleapis.com/v1/projects%2Fsanguine-rhythm-105020%2Fsubscriptions%2Frust_authd_sub_1?alt=json` produces a 404 error, but it works with `https://pubsub.googleapis.com/v1/projects/sanguine-rhythm-105020/subscriptions/rust_authd_sub_1?alt=json`.

The general change is basically not encoding `/` in URLs, and the `DEFAULT_ENCODE_SET` does that.

(also, I whaven't regenerated the APIs as I didn't feel like hunting after the issue that stops python from working right on my machine...)

### DIFF
--- a/gen/cloudbilling1/src/lib.rs.in
+++ b/gen/cloudbilling1/src/lib.rs.in
@@ -480,7 +480,7 @@ impl<'a, C, A> BillingAccountGetCall<'a, C, A> where C: BorrowMut<hyper::Client>
 
     /// Perform the operation you have build so far.
     pub fn doit(mut self) -> Result<(hyper::client::Response, BillingAccount)> {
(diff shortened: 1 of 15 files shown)


dermesser:
(got my python fixed)

Byron:
Awesome - thanks for the fix !
By the way, to me it's a bit distracting to have the regenerated APIs in the same PR. Would it be enough to point out that you would like to have it regenerated, so I can just do it separately ?

dermesser:
Yes! I wasn't sure which way you prefer, but I'll remember now.

PR #155, opened 2016-09-21
fix(api types): Add an unused field to empty API types.

dermesser:
Null structs (`struct Foo;`) cause the following error when trying to
deserialize an empty JSON object `{}` into them:

```
JsonDecodeError("{}\n", Syntax(InvalidType(Map), 1, 1))`
```
string representation:

```
invalid type: map at line 1 column 1: {}
```
The optional struct member prevents this error.


Byron:
Thanks a lot for finding the issue, and fixing it right away ! By now you must have some understanding on how the `mako` code-generation works. Good to have some evidence it is indeed usable by contributors :).

PR #157, opened 2016-11-07
Fix: multirust is deprecated - use rustup :)

Morsicus:


Byron-TW:
Thanks a bunch :) - Good catch!

PR #158, opened 2016-12-15
fix(discovery): handle discovery urls with $

compressed:
Some google discovery URLs contain `$discovery` or other variants,
causing the calls to wget to interpret `$d` as an environment variable
instead of a literal. An example is:
`https://logging.googleapis.com/$discovery/rest?version=v2`.

To fix this, the `$` has been escaped so that wget fetches the URL as
expected.

### DIFF
--- a/src/mako/deps.mako
+++ b/src/mako/deps.mako
@@ -179,11 +179,14 @@ help${agsuffix}:
 	target = target_dir + '/' + name + '-api.json'
 	## assure the target never actually exists to force him to wget whenver we ask !
 	fake_target = target + '-force'
+	## Some service urls have $ in them. This may cause the console to treat them as env vars.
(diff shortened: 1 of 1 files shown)


compressed:
Also if you accept this, it would be great if you can update the APIs since this change adds some crates that were missing before (logging, etc.) due to this bug.

Byron:
Thanks a lot for your contribution! I will now pull in the latest JSONs and publish the respective crates.

compressed:
Ha, yeah not sure why they chose to use a `$` in the some of the URLs.

Thanks for updating and for all your work here with this crate, I appreciate it!

Byron:
Thank you :)! And I sure hope it will be useful to you!
If you find anything, just let me know and I will try to be helpful.

PR #160, opened 2017-01-31
chore(deps): Update hyper for generated crates to ^0.10

dermesser:
You guessed it, again related to the openssl upgrade in dermesser/yup-oauth2#51. As long as the API crates depend on openssl 0.7 via hyper 0.9, any client using the APIs won't build, because yup-oauth2 pulls in openssl 0.9 (...because openssl 0.7 doesn't build on Mac). It's one big yak to shave, and I'm sorry for all the trouble :(

Please regenerate and publish afterwards, if possible.

### DIFF
--- a/src/mako/Cargo.toml.mako
+++ b/src/mako/Cargo.toml.mako
@@ -26,7 +26,7 @@ name = "${util.program_name()}"
 % endif
 
 [dependencies]
-hyper = "^ 0.9"
+hyper = "^ 0.10"
 ## Must match the one hyper uses, otherwise there are duplicate similarly named `Mime` structs
 mime = "^ 0.2.0"
 serde = "^ 0.8"


Byron:
Thanks a lot! I actually managed to get rid of openssl entirely! What do you think of doing that for yup-oauth2 as well?

dermesser:
at the moment, yup-oauth2 relies on OpenSSL not only via hyper, but also for JWT signing (https://github.com/dermesser/yup-oauth2/blob/master/src/service_account.rs). When I wrote it, the other JWT crates were kind of murky, so I had to implement the signing using OpenSSL directly.

Those operations are possibly provided by rustls as well; I just haven't looked at it yet. In that case, it'd certainly be possible and advisable.

Byron:
Alright, I will make the new release in the course of the day, and hope you will have some time to look into it - would be awesome to have a pure-rust implementation of the google-bindings :)!

dermesser:
I'll try it on the weekend, and I agree that it'd be pretty nice!

PR #167, opened 2017-03-31
Updated index.html documentation template

### LINKED ISSUE #166: Malformed documentation
It looks like the html for http://byron.github.io/google-apis-rs/ got a little messed up. I've attached screenshots of the offending sections.

Would it be better to use a table instead of a bunch of spans and `<br>` tags? If you can point me at whatever file you're using as the page template, I'd be happy to fix it up.

screenshot_2017-03-31_211357
screenshot_2017-03-31_211413

Michael-F-Bryan:
I believe I've figured out why the lines were getting messed up. Basically, some API names were exceeding the 300px allocated to them by `.lib`, so they were being pushed down to a new line. This then messed everything up for the line below so that it starts 300px in.

I converted from the old span + br way of formatting to using tables to keep everything laid out. I hope you don't mind, but I added a link tag to pull in [Bootstrap]'s css from their CDN to make things look pretty.

At the moment the page looks something like this:
screenshot_2017-04-01_024532

(Fixes #166)

[Bootstrap]: http://getbootstrap.com/getting-started/#download-cdn

### DIFF
--- a/src/mako/index.html.mako
+++ b/src/mako/index.html.mako
@@ -69,79 +69,75 @@ function onCopy(e) {
 	<title>${title}</title>
 </head>
 <body>
-    <div class="container">
-<h1>${title}</h1>
-<table class="table table-hover">
+  <div class="container">
+  <h1>${title}</h1>
+  <table class="table table-hover">
     <thead>
-        <tr>
-            <th>API Name</th>
-            <th>API Docs</th>
-            <th>CLI Docs</th>
-            <th>Install</th>
-        </tr>
+      <tr>
+        <th>API Name</th>
+        <th>API Docs</th>
+        <th>CLI Docs</th>
+        <th>Install</th>
+      </tr>
     </thead>
     <tbody>
     % for name in sorted(api.list.keys()):
-        % if name in api.blacklist:
-            <% continue %>\
-        % endif
-        % for version in api.list[name]:
-        <tr>
-            <% 
-                # We know type_names is just ["api", "cli"]
-                #type_names = tc.keys()
-                type_names = ["api", "cli"]
-                with open(api_json_path(directories.api_base, name, version)) as fp:
-                    metadata = json.load(fp)
+      % if name in api.blacklist:
+        <% continue %>\
+      % endif
+    % for version in api.list[name]:
+      <tr>
+        <% 
+            # We know type_names is just ["api", "cli"]
+            #type_names = tc.keys()
+            type_names = ["api", "cli"]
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
There's still a small issue where the "+..." part of the API links gets set to "+000000", but once I get that fixed and check out the page on a couple other browsers and my phone it should be good to merge.

src/mako/index.html.mako
```diff
@@ -80,44 +68,77 @@ function onCopy(e) {
...
+      % endif
+    % for version in api.list[name]:
+      <tr>
         <% 
-            type_names = tc.keys()
-            with open(api_json_path(directories.api_base, an, v)) as fp:
-              api_data = json.load(fp)
+            # We know type_names is just ["api", "cli"]
+            #type_names = tc.keys()
+            type_names = ["api", "cli"]
```
Byron: I don't want to be nitpicky, but what is the reason for not procedurally obtaining information that is available already, but to hardcode it? This will be a bug once there is another class of derivative, like `gui` maybe?

Byron: Maybe it's because the code is already making assumptions? I really don't know it enough anymore to know for sure, but that would certainly be a valid reason and fine as well :). After all adding a new derivative like `gui` is a lot of work, and doesn't happen every day.

Michael-F-Bryan: My thoughts were that it's not very likely that another class of derivative will be developed and that by doing it dynamically it was adding a lot of complicated logic to the template. Usually I prefer to keep as much logic out of the template as possible because that tends to make things simpler and more maintainable.

If you want, I could add something like `assert type_names == tc.keys()` so that if things ever change in the future we're notified straight away that the template needs to be updated. How does that sound?

Michael-F-Bryan: It looks like the original template also implicitly assumed there would only ever be two variants (`api` and `cli`) because there's an if/else statement which says if it's `api` to print a link to the api docs and crates.io, otherwise print the button to install the cli tool.

So chances are adding another variant would have ended up with introducing bugs anyway.

Byron:
Sounds great:)! Please feel free to remove the commented code also.
Maybe even skip the assertion if .keys() order is undefined.
I shall merge this later, possibly even making said adjustments myself.
Thanks so much!

Michael-F-Bryan:
Awesome, this should be ready to merge.

Byron:
Awesome! Thanks a lot for helping me out! 
Later today I shall update the documentation to get your work out there :)!

Byron:
And it's done! Looks great!

PR #168, opened 2017-05-07
feat(hub): Allow overriding rootUrl and baseUrl

LegNeato:
Allow the hub to override `rootUrl` and `baseUrl` for the service.
This is useful for pointing at localhost for testing or a proxy.

### DIFF
--- a/gen/adexchangebuyer1d3/src/lib.rs
+++ b/gen/adexchangebuyer1d3/src/lib.rs
@@ -191,7 +191,7 @@
 
 // Unused attributes happen thanks to defined, but unused structures
 // We don't warn about this, as depending on the API, some data structures or facilities are never used.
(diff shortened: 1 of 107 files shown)


LegNeato:
(test failures on nightly look unrelated)

Byron:
Sorry for letting you hang for so long, your contribution is very much appreciated!
I will get to it next week I think.

LegNeato:
No worries, take your time...I definitely do not consider 2 days over a weekend a long time 😃

Byron:
@LegNeato Thanks for the contribution! A great idea to add this capability :)!
There is only one thing I improved, which is the usage of `std::mem::replace` instead of the usual swap code `t=x;x=y;t` - the former avoid borrow checker issues.

PR #177, opened 2017-08-23
Strip leading slashes from URL-s

### LINKED ISSUE #171: unclear how to create a folder with google_drive2
I've been assuming that the correct call for this is Drive::files().insert(). However, when using this to "upload" an empty file whose mime type is application/vnd.google-apps.folder, I get a 500 error in response.

This is an excerpt of the code that I'm using unsuccessfully:

                let fm = Mime(TopLevel::Application,
                    SubLevel::Ext("vnd.google-apps.folder".to_string()),
                    vec![]);

                let emptybuf: [u8; 0] = []; 
                let empty_stream = Cursor::new(&emptybuf);
                let mut folder_info: drive2::File = Default::default();
                folder_info.title = Some(drive_folder_name);
                folder_info.mime_type = Some(fm.to_string());
                hub.files()
                    .insert(folder_info)
                    .add_scope(drive2::Scope::File)
                    .param("fields", "id")
                    .upload(empty_stream, fm) 

Other Google Drive APIs seem to have an `execute` method to perform the file insert operation, which feels more correct in the case of creating a folder, rather than uploading an empty file. However, I'm not clear how else to perform this using this Rust API, which only provides `upload` methods to perform the files.insert request.

https://developers.google.com/drive/v2/web/folder

cristicbz:
Some URLs had leading slashes which, when concatenated with the root url introduced in #168 would result in double slashes after the hostname.

(before #168 the `os.path.join` call would get rid of them, but the Rust code just concatenates)
 
I think this fixes #171 (at least the 500 error they're getting). In our code, we were seeing this 500 error when using any `insert` with `google_storage1`, so I think it's a pretty critical fix.

I tired to meet the commit style of the project, but let me know if you'd like it done in a different way, happy to amend the history.

Thanks!

### DIFF
--- a/gen/adexchangebuyer1d3/src/lib.rs
+++ b/gen/adexchangebuyer1d3/src/lib.rs
@@ -2067,13 +2067,20 @@ impl<'a, C, A> BillingInfoGetCall<'a, C, A> where C: BorrowMut<hyper::Client>, A
     ///
     /// The `scope` will be added to a set of scopes. This is important as one can maintain access
     /// tokens for more than one scope.
+    /// If `None` is specified, then all scopes will be removed and no default scope will be used either.
+    /// In that case, you have to specify your API-key using the `key` parameter (see the `param()`
+    /// function for details).
     ///
     /// Usually there is more than one suitable scope to authorize an operation, some of which may
     /// encompass more rights than others. For example, for listing resources, a *read-only* scope will be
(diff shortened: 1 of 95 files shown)


Byron:
My apologies for the delay, thanks so much for the fix! I will try to get the fix released in the next days.

PR #183, opened 2018-01-26
Fix for util.py to allow for sheets api to be built

faddi:
Sheets api has a 'values' key in resources.spreadsheets.resources which collided with activities.values(). Using iteritems to iterate instead.


Byron:
Thanks so much for figuring out a fix! I have also published the `sheets` crate so it should be easy to use: https://crates.io/crates/google-sheets4-cli .

PR #184, opened 2018-01-31
Enable compute1

mdaffin:
Removed `compute1` from the blacklist and `compute1-cli` builds and runs without issue on latest rust (1.20.0) and `compute1` tests run without issue on the rust version 1.17.0-1.20.0.

Is there any other reason this cannot be published?


mdaffin:
Note, I was unsure if the comment above the entry in the blacklist was just for the compute, or was for the services below it as well.

Byron-TW:
Thanks a lot! Thanks to the recent fix in the generator, and maybe differences in Rust, it may be that other APIs also started working.
For now I will just release the new ones as they come in.

Byron-TW:
`compute-1` is now available on crates.io, and the CLI docs are available on the google-apis-rs site, too.

mdaffin:
Thanks. This was the only API I needed for now. But if you want I can go through the others and see if they build and create a pull request for each one that does or one pull request for all that do.

Byron:
That would be great!
For me it’s easier to have less PRs, but I happily go with what’s easiest
for you.

PR #185, opened 2018-02-01
Enable working apis

mdaffin:
Removes from the blacklist all of the APIs that successfully build on the latest version of rust (1.23.0) and adds all of the generated files for these APIs.

It looks like some of the tests are failing - but they are failing for some APIs that I have not touched (such as `partners2`) and they are failing on master.

### DIFF
--- a/etc/api/shared.yaml
+++ b/etc/api/shared.yaml
@@ -12,30 +12,15 @@ api:
     - kgsearch
     - mapsengine
     - firebaserules
-    - runtimeconfig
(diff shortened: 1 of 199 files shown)


Byron-TW:
Wow, this is rich!
There aren't that many tests actually, it just tries to build a few APIs on CI and that's it. It seems for the most part, they are working though.

I will publish the missing APIs and update the docs accordingly.
Thanks a lot for contributing :)!

PR #188, opened 2018-04-01
chore(deps): Pin hyper and mime crates to fix compilation.

dermesser:


dermesser:
Not really, I was just compiling it again, and it failed on these. I fixed them because unpinned crates are an easy source of problems :)

as to why I was compiling it... :P I was experimenting to find out how hard it is to convert the current structure to the latest hotness, i.e. futures (which is provided by appropriately new versions of hyper). I'm not very happy with the futures+tokio situation right now (it's an actively developed mess, so quite unstable, which is why we shouldn't invest much time for now).

Anyway, the tl;dr is that the conversion is not trivial, but probably doable; if you'd be interested as well I can imagine putting some work into it when the outside situation has stabilized.

Byron:
That's great to hear! I think very similarly!
My implicit plan is to
* get rid of the python template engine in favor of something rusty
   * liquid templates can do it already, and I would prefer a more test-driven approach to things. As liquid doesn't support anything fancy, one would have to take the json API declarations and process them (in Rust) to contain all derivations and information we would need to generate the output files. That way, the templates don't contain too much logic, as is currently the case.
* rethink the design of the library.
  * Lessons learned here is that 'writing everything out' is good for huge binaries and long compile times. I think you can have type-safety and no repetition by using trait objects and by leveraging the 'Serialize' trait on structs. That way it's the best of both worlds, with relatively small binaries, and fast compile times.
* use stable tokio 1.0.
  * By that time it should be easy to be async first, knowing that users of the library can easily await on the futures it produces, making it sync as required.
* use the latest greatest yup-oauth2
  * ideally we also provide all the auth-features people desire by that time :).
* assure CLIs stay a first-class output of this project.
  * it should be easy to use google from the command-line, which is the original motivation for this project.

I think the overarching goal should be to be the definitive, _go to_ implementation of google api clients in the Rust ecosystem.

PR #201, opened 2018-09-27
Photoslibrary

### LINKED ISSUE #198: Add photos API

TheBiggerGuy:
The photoslibrary API is not listed however is still available.
This adds a method of manually adding APIs to the shared.yaml and adds the photoslibary API info.

Fixes #198

### DIFF
--- a/etc/api/api-list.yaml
+++ b/etc/api/api-list.yaml
@@ -200,6 +200,8 @@ api:
     - v2
     people:
     - v1
+    photoslibrary:
+    - v1
     playcustomapp:
     - v1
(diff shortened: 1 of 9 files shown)


TheBiggerGuy:
So I found the first issue with the API.
In the [albums](https://developers.google.com/photos/library/reference/rest/v1/albums) type the _mediaItemsCount_ property is of type _int64_ but returned as a **string** version of a number.

I have yet to understand enough to work out how to map this type correctly.

Byron:
I am looking forward to see this become mergeable :)!

PR #202, opened 2018-09-27
Removed apis

TheBiggerGuy:
Wget leaves empty files when getting a 404. This causes JSON parse errors later on instead of the expected file not found message.

This is caused by some API versions been listed but not existing. Therefore I have added new code to be able to blacklist versions of APIs.


Byron:
Thanks a lot, @TheBiggerGuy ! It's looking great!

PR #204, opened 2018-10-09
Update hyper

paperplanes53:
This is my first attempt at incorporating the hyper 0.12.11 features.  I am using hyper-tls for a project, however another connector could be used.


Byron:
Thanks for giving this a serious shot! Please let me know if there is anything I can help you with, even though by now I don't know my ways in the project anymore either.

One recommendation I can provide you with right now: One should not checking changes to the generated output, as it is likely to conflict with master soon as the APIs are re-generated from time to time. For local testing, however, generating them is the only way to verify changes.

PR #207, opened 2018-10-16
RFC: Python testing

TheBiggerGuy:
**Request for comments** on the style of Python testing

This is a simple version with Make triggering PyTest.

My current view is ether keeping it simple or try and introduce a more Pythonic flow (i.e tox, setup.py...).
Simple system:
* Easy to understand
* No too Pythonic for the Rust people and not too Rust like for the Python people.
* Get complicated quick if expanded with PyLint or multiple Python version tests

Pythonic flow:
* Not reinventing Python testing with Make
* Help later in a Py3 migration
* Easier to introduce Lint, code style and other tooling


Makefile
```diff
@@ -81,6 +82,9 @@ license: LICENSE.md
 
 regen-apis: | clean-all-api clean-all-cli gen-all-api gen-all-cli license
 
+gen-test: $(PYTHON)
```
Byron: Firs of all, I love all of this!
When reading the make-target for the first time, I thought it would be generating tests, even though in fact it is running them.
Renaming the target would do the job - this is just a suggestion:
```suggestion
test: $(PYTHON)
```

Byron:
Oh, and by the way: please feel free to merge this one yourself :).

Byron:
That sounds reasonable! If that changes at some point, it's no problem to rename it, either :).

TheBiggerGuy:
So I'll wire it into Travis and ensure that goes green before merging.

Byron:
Alright! I will help in case you don't have enough privileges. If you provide a `.travis.yml`, I should be able to do the rest if needed.

TheBiggerGuy:
So I am leaning on wanting to go with
```yaml
script:
  - make test-gen
  - make gen-all-cli cargo-api ARGS=test
  - make cargo-api ARGS=doc
  - "if [[ $TRAVIS_RUST_VERSION = nightly ]]; then cargo test; fi"
```
As running tests on the code that builds the rest before it it built seems right. I would like to see the Rust code tests run with just a ```make test``` as the _ARGS=test_ seems a bit awkward.

TheBiggerGuy:
OK. Took a minute but I've worked out why Travis was not working. I broke it in my last pull :(

I'm waiting for https://travis-ci.org/TheBiggerGuy/google-apis-rs/builds/444366490 to finish and then I will pull it into master.

PR #209, opened 2018-10-21
Fix Travis Yaml file

TheBiggerGuy:
The Travis Yaml file was invalid after #205

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,10 +27,11 @@ script:
 env:
   global:
     - PYENV_VERSION: 2.7.14
+    - PYENV_ROOT: $HOME/.pyenv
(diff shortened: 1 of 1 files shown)


TheBiggerGuy:
Rust stable had a runtime of 6m 33 seconds. https://travis-ci.org/Byron/google-apis-rs/jobs/444372166
I am going to restart it a check the cache works, flush the cache and repeat, make sure it is working.

TheBiggerGuy:
Cached build took 3m 5 seconds so 2x speed improvement. Unfortunately the Python (pip/wheel) cache does not work due to the virtualenv config.

Byron:
Pretty neat already! Can't wait for it to land in master :).

TheBiggerGuy:
So without a working `.pyenv_cache` the OSX builds are ~3mins slower. To be clear this is a cache of the Python binary, this is due to an old version of Python on Travis, and not the Python VirtualEnv or Pip/Wheel cache.

PR #214, opened 2018-12-07
docs(sheets): fix build errors

### LINKED ISSUE #213: Docs: The requested resource does not exist
I want to use this crate to append rows to a google sheet, but where can I find the docs?
Here it says "The requested resource does not exist":
https://docs.rs/google-sheets4/1.0.8+20181010/google_sheets4/

IslandUsurper:
Rustdoc assumes indented code is Rust code, but when it isn't, the docs
fail to build. The Sheets API has Java, Objective-C, and JavaScript
examples in its descripiton. Wrap these examples in language-tagged
triple-backticks to prevent rustdoc from trying to compile them.

Fixes #213.


Byron:
Thanks a lot for the fix!
I am a bit worried that next time someone calls `make update-json`, the fix will be overwritten. 

Something I can imagine is to keep the patch for this fix around and have the system apply it automatically in case a patch file exists.

Let's see how it will go.

PR #217, opened 2019-02-15
fix(rust): teach remove_json_null_values about arrays

cpick:
change `remove_json_null_values()` to properly remove nulls from and recurse into arrays google_firestore1_beta1's `CommitRequest` contains an array of `Write` objects which can ultimately contain `Value` members that need to have nulls removed to avoid sending multiple types of values which generates a 400 response

fixes calls to google_firestore1_beta1's `hub.projects().databases_documents_commit()`


Byron:
Sorry for the late reply and thanks a lot for the PR! It's much appreciated! I am preparing a re-release of the existing crates.

PR #219, opened 2019-02-28
Pin yup-oauth2 version to 1.x

TheBiggerGuy:
This ensures the yup-oauth2 version stays with the 1.x major version
allowing the yup-oauth2 project to make a breaking change via a 2.x
major version.

Ref: https://github.com/dermesser/yup-oauth2/pull/86


Byron:
Thanks a lot. I will try to re-publish these crates once the PRs are merged.

PR #222, opened 2019-04-04
April refresh

TheBiggerGuy:
* **Sort JSON files to minimise git diffs**
Sort and format JSON files to remove/minimise any diffs between `update-json` runs. This also acts as a JSON validator to error when the downloaded JSON file has been truncated.

* **Add logging and switch to curl**
I am getting many truncated JSON file that then do not parse. More from wget to curl to see if this helps.

* **Update JSON and regen all APIs**


Byron:
Thanks a lot for your contribution. Let me see if I can manage to publish all of these now :D!

PR #223, opened 2019-04-08
docs: add lang attribute to docs index

amy-keibler:
Use the W3 standard "lang" attribute on the "html" tag to provide context for browsers and screen
readers.

Context: Chrome has a language heuristic that determines the page is in Portuguese. It displays the "Translate this page" dialog. Adding the "lang" attribute context should prevent this dialog.

Reference: https://www.w3.org/International/questions/qa-html-language-declarations


Byron:
Thanks a lot for the fix! I will hopefully get to do a re-release/update in the next days.

Issue #225, opened 2019-04-26
Possible API alternative


ggriffiniii:
First let me say thanks for generating all these API's, the rust ecosystem needs them and they provide a ton of value.

I have an idea and proof of concept for how to interact with google API's in rust that I think has a lot of benefits, and given that your libraries are the de facto way of using google API's in rust I wanted to run it by you in case you had any thoughts.

Most of the modern google API's provide a `fields` attribute on the request that controls which attributes you want populated in the response. You can see the google drive API for the fields attribute [here](https://developers.google.com/drive/api/v3/performance#partial-response), but most of the current API's have identical functionality.

Your currently generated API (along with API's in other languages) handle this by having a response struct that contains a bunch of optional fields. For example google_drive3::File currently has over 50 fields each of which is optional and some have nested sub-fields. The two big issues with this are that while it's type-safe the type itself does not represent what fields it effectively contains, and because the struct maintains every possible field it's larger in memory than all but the most extreme cases require.

As a user of the library I wished I could define my own struct that implemented serde::Deserialize and have the responses deserialized into my struct. This would allow me to express my data in a more rich way using the type system and use the powerful features of serde to deserialize into appropriate types like chrono::DateTime where appropriate. This is pretty straightforward, but now the `fields` attribute on the request had to be crafted to match the fields specified on my struct. To solve this I created a procedural macro that can automatically derive a `field_selector`. When the field_selector is provided in the request it will request all the fields in the struct. I'm pretty satisfied with the way it ended up and thought it might make for an interesting approach for the generated API's you provide.

My proof of concept can be found here:

- https://github.com/ggriffiniii/google_api3
  - That repository defines the FieldSelector trait along with a proc macro to derive it on structs. Any struct that implements FieldSelector can call T::field_selector() to generate a selector for the `fields` attribute.
- https://github.com/ggriffiniii/simple_drive_api
  - This repository has a barebones implementation of a google drive client that only lists files. It's intentionally stripped of a lot of functionality to highlight how FieldSelector can be used. The concept of how FieldSelector is used can be applied to the vast majority of modern google api's.
  - In particular https://github.com/ggriffiniii/simple_drive_api/blob/master/examples/list_files.rs has an example of how a user of such a library would operate. Note that it's doing 2 different file list operations and each is returning a different struct with different fields populated.

[...]

PR #231, opened 2019-07-12
Allow deserializing into custom responses

ggriffiniii:
This is a pretty quick implementation that allows deserializing into custom structs as outlined in issue #225.

This does the simplest and most straightforward thing. Wherever there is a `doit` method there is now an accompanying `doit_into` that has a generic return type that needs to implement DeserializeOwned + Default + FieldSelector.

All the existing custom types have been updated to implement FieldSelector so they can be used with the `doit_into` methods and the `doit` methods are implemented as calls to `doit_into`.

I think there remain some rough edges, but in general I think this works pretty well. I didn't want to publish any new crates until after the PR has been reviewed so the dependencies are currently specified with path specifications. These will need to be removed and updated to proper crate names when publishing. The documentation also needs to get updated, but I wanted to send this for review before polishing up those details.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 # This library is just to try out the code that should ultimately go into the code generator !
 
 [workspace]
-members = ["google_field_selector", "google_field_selector_derive", "gen/*"]
+members = ["google_field_selector", "google_field_selector_derive", "gen/*", "src/rust/preproc"]
 
 [package]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,9 @@
 # DO NOT PUBLISH
 # This library is just to try out the code that should ultimately go into the code generator !
+
+[workspace]
+members = ["google_field_selector", "google_field_selector_derive", "gen/*"]
+
 [package]
 
 name = "cmn"
--- a/google_field_selector/Cargo.toml
+++ b/google_field_selector/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "google_field_selector"
+version = "0.1.0"
+authors = ["Sebastian Thiel <byronimo@gmail.com>"]
+edition = "2018"
+
+[dependencies]
+google_field_selector_derive = { version = "0.1.0", path = "../google_field_selector_derive" }
+
+[dev-dependencies]
+chrono = { version = "0.4", features = ["serde"] }
+serde = { version = "1", features = ["derive"] }
--- a/google_field_selector/src/lib.rs
+++ b/google_field_selector/src/lib.rs
@@ -1,0 +1,143 @@
+pub use google_field_selector_derive::FieldSelector;
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet, LinkedList, VecDeque};
+/// FieldSelector provides a google api compatible field selector. This value can
+/// be provided in the "fields" attribute of many google api's to request which
+/// subfields to include in a a response. Implementations of FieldSelector will
+/// typically be automatically generated from a procedural macro using
+/// `#[derive(FieldSelector)]`.
+pub trait FieldSelector {
+    fn field_selector() -> String {
+        let mut selector = String::new();
+        Self::field_selector_with_ident("", &mut selector);
+        selector
+    }
+
+    fn field_selector_with_ident(ident: &str, selector: &mut String);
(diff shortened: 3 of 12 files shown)


Byron:
Thanks a lot, I am very excited to see this PR go into a new release.

While trying it locally, from scratch, I ran into this cargo related problem, though, maybe you could have a look, maybe it's expected in the current state.

```sh
make regen-apis
```

Causes this output:
```
[...]
error: current package believes it's in a workspace when it's not:
current:   /Users/sthiel/dev/google-apis-rs/src/rust/preproc/Cargo.toml
workspace: /Users/sthiel/dev/google-apis-rs/Cargo.toml

this may be fixable by adding `src/rust/preproc` to the `workspace.members` array of the manifest located at: /Users/sthiel/dev/google-apis-rs/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
[...]
```

Once that is fixed, I guess it's fair to add examples on how to use this new capability - but it's something for the future.
Thanks a lot!

ggriffiniii:
Interesting. I'm not sure why I didn't see that error. When I do a make regen-apis I don't get that error, but I do see it when I `cd src/rust/preproc; cargo build`. I'm not sure what the preproc directory does, but maybe it's not always built when you do a regen-apis? 

Either way there are a couple ways to fix this. I meant to mention that I wasn't sure how you typically deal with multiple crates in a single repository. I did the workspace changes just so that I could get things to work in my repository with the commands and system I was used to, but you can take them or leave them (or tell me to undo them on this PR). It should be pretty simple to add src/rust/preproc to the workspace.members list in the root Cargo.toml. I've done that in this PR.

Byron:
Thanks a lot - it works!
Now that I am looking at it, it seemed cargo already proposed the solution, and I don't quite know what I was thinking when bouncing this back here.

Looking at the code, I think I understand how it all comes together, and am looking forward to one day implement my own (proc-)macros :D!
One thing puzzles me:
```rust
 let fields = T::field_selector();
     if !fields.is_empty() {
         params.push(("fields", T::field_selector()));
     }
```
Why is `T::field_selector()` called twice, can't `fields` be used instead?

On another note: a lot of changes are inbound for me, which actually means that I will have time to fully dedicate myself to this project once again, with the goal to essentially re-implement everything in Rust and provide a fully `async` API. This PR or a version of the field selector will be a part of that, for sure, while giving ample opportunity to do even nicer things :).

Also CC @dermesser , as `yup-oauth2` already supports `async`.

ggriffiniii:
Calling T::field_selector multiple times was an oversight after refactoring. You're correct that it's best to only call it once.

Regarding re-implementing everything in pure rust with an async api let let me know if you would like any help with that. I've actually been going down that road on my own and wasn't really sure what I was going to do once I got things into a more finished state. I would be more than happy to share what I have with you and collaborate on the next iteration of your work if you're interested.

Byron:
Yes! Let's work on this together :)! Actually starting now, I could work full-time on this for a few weeks at least, which may be enough kick it off as a proof of concept at least.

The rough sketch I have in my mind involves creating a new repository to start fresh, this one is already ~150MB, packed.

As for this PR, do you think you could leave some usage example, at least in the main README, so we have something to refer to? That could conclude the active development on the sync version, alongside a last release.

What do you think?

ggriffiniii:
Actually, if there's going to be work on a new version I'm inclined to hold off on integrating this into the current version in favor of making it integrate cleanly into the new version. I'm not overly confident in the hacks I made to the python templates especially given the number of API's that are generated and the edge cases it's surely going to encounter.

In terms of the next version I'm fine starting with a blank slate repository. I've given you access to a WIP repository where I've started down the path of a rust only generator. It's far from complete and sorely lacking in documentation, but there's probably some pieces worth building off of. It's https://github.com/ggriffiniii/google-apis I'm in a rush right now, but I'll try and followup with a summary of what currently exists in that repository so we can see how close our visions line up.

Byron:
@ggriffiniii That sounds sane! Let's not rush things unnecessarily.

I will try to write down my thoughts about the goals of rewrite into the README.md of the project you invited me to, and will also take a look at what's already there :).

Byron:
Closed to indicate no further work will be done here.

PR #250, opened 2020-04-13
Fix spelling of "encodable"

relud:


Byron:
Thanks a lot for the fix!

I have extracted the fix within the Mako file and pushed it separately, to avoid regenerating all code before releases. That way, this repository stays a little smaller. Thanks for your understanding.

PR #254, opened 2020-05-29
Extend attempted error message parsing to also cover ErrorResponse objects

bjornax:
Fixes errors in e.g. drive3 API not getting passed to delegate error handlers.
An ErrorResponse is basically a ServerError wrapped in an "error" object. See e.g. https://developers.google.com/drive/api/v3/handle-errors

A few other APIs that I randomly sampled could probably be helped by this also, e.g. the youtube API.


Byron:
Awesome, great work!

Thanks so much!

Issue #232, opened 2019-07-21
How do I rename a file? (google-drive3)


dvaerum:
Hope someone can help, I am trying to rename a file on google drive using this library, but I simple cannot get it to work.

From what I can read on in the documentation here and the official one from Google it should be something like this, but the fucktion doit() is private and cannot only called by update() and upload_resumable(), but I don't want to overwrite the file, I just want to update its name.
```Rust
type HubType = DriveHub<Client, Authenticator<DefaultAuthenticatorDelegate, DiskTokenStorage, Client>>;

fn rename(hub: &HubType, file_id: &str, new_name: String) {
    let mut file = File::default();
    file.name = Some(new_name);

    let result = hub.files()
        .update(file, file_id)
        .supports_all_drives(true)
        .doit();

    info!("Rename - {:?}", result)
}
```

I hope someone can help, because from what I can see this will also be a problem if I want to move the file.

dvaerum:
So I got it to work by taking the function [doit_without_upload](https://docs.rs/google-drive3-fork/1.0.10/src/google_drive3_fork/lib.rs.html#3428-3585) which I found in a forked version [google_drive3_fork](https://github.com/Byron/google-apis-rs/tree/master/gen/drive3).
I had to make some small changes for it to work, but after that it seems to work :D

I have created a patch if anyone should be interested [doit_without_upload.patch.txt](https://github.com/Byron/google-apis-rs/files/3439485/doit_without_upload.patch.txt). I was trying to figure out there in the generator I should make changes in order to get the doit_without_upload function included, but I am not able to find it.

Can anyone give me a pointer?

Byron:
Sorry for the late reply, and great to hear that you found code that works. The next step could be to apply the change to the generator, which [can be found here](https://github.com/Byron/google-apis-rs/blob/1565e040c5b66fc03acec5eb9a8c2a5d570f3056/src/mako/api/lib/mbuild.mako#L503).

I am looking forward to a PR in case you find the time. In any case, I will be looking at the issues that came up while using various APIs while re-writing the crate to make use of `async`, and contain all lessons learned so far.

dvaerum:
Hey @Byron it has been sometime since I look at this, but today I gave it another try. I was digging around in your generator and it was a little overwhelming, so I decided to watch your "Developer Diary" videos and it gave me some inside.
[...]

PR #255, opened 2020-06-03
Doit without upload

dvaerum:
Related to: https://github.com/Byron/google-apis-rs/issues/232

Hey @Byron here is your long awaited pull request :wink: and after I played around with your code generator it turned out that I really didn't need to change that much to get what I wanted, which is awesome :+1: 

I have split the pull request up in 3 parts.
- [1st commit](https://github.com/dvaerum/google-apis-rs/commit/1cc94d5281952c9b642be332254364ebe7e85389)  
I re-generates all the code because it created changes which was not commit and by commit these changes, it makes it a lot easier to see what I have changed and was impact it has.
- [2nd commit](https://github.com/dvaerum/google-apis-rs/commit/041f8951e3409e26c511350a73c1aef0b26d307e)
Basically, just added 4 `if` statements, your generator takes care of the rest :smiley: 
- [3rd commit](https://github.com/dvaerum/google-apis-rs/commit/8f7ed2e5e9621e7e9f824f397a68dcf9cf6df820)
The change there happens to the codebase after I rebuild everything, just to confirm that I didn't break/interfere with other apis/libs

The code generates a doit_without_upload function for both `drive2` and `drive3`, I only use `drive3` but I added it to `drive2` because after reading the docs and looking at the code, it was the same as `drive3` just with a little difference in naming in some of the `struct`s.

**Note:** I can see that some other lib are not compiling when I tried to compile everything, but I can see that my changes to the code generator don't touch these libraries

### DIFF
--- a/etc/api/drive/v2/drive-api_overrides.yaml
+++ b/etc/api/drive/v2/drive-api_overrides.yaml
@@ -1,2 +1,3 @@
 api:
   credentials: "{\"installed\":{\"auth_uri\":\"https://accounts.google.com/o/oauth2/auth\",\"client_secret\":\"De0ub0IbWruJbBXUyseFYvZ-\",\"token_uri\":\"https://accounts.google.com/o/oauth2/token\",\"client_email\":\"\",\"redirect_uris\":[\"urn:ietf:wg:oauth:2.0:oob\",\"oob\"],\"client_x509_cert_url\":\"\",\"client_id\":\"276875258587-5gbp23a7aqnrl6p06c0jt5fskuktactq.apps.googleusercontent.com\",\"auth_provider_x509_cert_url\":\"https://www.googleapis.com/oauth2/v1/certs\"}}"
+  no_upload_prefix: FileUpdateCall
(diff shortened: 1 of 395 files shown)


Byron:
Hi @dvaerum , it's great to see this wonderful PR!

While studying it I actually understood the code-generation a little better myself, once again. I actually kind-of like it, allowing to build a model by merging multiple files into one, which then is used by a templating engine.

This also made me realize that instead of hard-coding a special case like that, a proof of concept, we could implement it using per-API overrides.
And because it was quite straightforward and a nice exercise, I have done just that and merged everything back into master.

Is there anything else that you think should be done now?

dvaerum:
> Hi @dvaerum , it's great to see this wonderful PR!

Thanks, I am happy that you like it :D

> While studying it I actually understood the code-generation a little better myself, once again. I actually kind-of like it, allowing to build a model by merging multiple files into one, which then is used by a templating engine.

Good to hear that you also got something out of it :smiley: I was trying a few different thinks and while playing around, just found that the code was already there, I just needed to enable it. It was a really awesome moment and it made me even more impress with your code generator. 

> This also made me realize that instead of hard-coding a special case like that, a proof of concept, we could implement it using per-API overrides.
> And because it was quite straightforward and a nice exercise, I have done just that and merged everything back into master.

Yes, I saw that commit, I just didn't know that this was an option, but I really likes it :+1: 
https://github.com/Byron/google-apis-rs/commit/5e5327f3f789db0e451de39e0d14c708426be24e
 
> Is there anything else that you think should be done now?
No, I got what I wanted :wink:

PR #256, opened 2020-07-08
Fix '-force' target generation for JSON schemata.

joell:
In some environments, a variable-existence test in a template was
yielding a false negative which resulted in the template not generating
'-force' targets for API JSON files, even when the `FETCH_APIS`
environment variable was set during template rendering.

This commit adjusts the way the test for the variable presence is done
such that the '-force' targets now render when they should.

### DIFF
--- a/src/mako/deps.mako
+++ b/src/mako/deps.mako
@@ -7,6 +7,7 @@
 	import urllib2
 	import json
 
+	apis = {}
 	api_info = []
 	doc_root = directories.output + '/' + directories.doc_subdir
 	doc_index = doc_root + '/index.html'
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot for the fix! 
I vaguely remember having fiddled with this portion of the code so might as well have broken it - apologies!

joell:
No apologies are necessary at all!  You are the one who's invested numerous hours into providing this code and framework as a service to the community; that's not something you need ever apologize for, even if you find an imperfection.  Instead, **THANK YOU** for all the work you've put into this which so many of us have been able to benefit from!

PR #257, opened 2020-07-08
Update the PubSub schema to include new API features.

joell:
Since the last time the PubSub v1 schema was fetched from Discovery in
April, Google has added new API features: namely, subscription filters
and retry policies.  This update to the JSON schema and its generated
code bring in support for these features.

### DIFF
--- a/etc/api/pubsub/v1/pubsub-api.json
+++ b/etc/api/pubsub/v1/pubsub-api.json
@@ -202,7 +202,7 @@
                             ], 
                             "parameters": {
                                 "options.requestedPolicyVersion": {
-                                    "description": "Optional. The policy format version to be returned.\n\nValid values are 0, 1, and 3. Requests specifying an invalid value will be\nrejected.\n\nRequests for policies with any conditional bindings must specify version 3.\nPolicies without any conditional bindings may specify any valid value or\nleave the field unset.", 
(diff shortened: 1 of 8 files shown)


Byron:
Thank you!
Maybe it's time for a complete re-build and re-publish. What do you think?

joell:
That sounds like a good idea to me.  Having the latest PubSub package in crates.io would make my life easier, and I'm sure others depending on other APIs might be in the same boat.

I actually took a quick stab at trying to re-fetch and re-generate all the APIs, but got some errors for some of them; it looked kind of like some APIs the build system was trying to pull didn't exist anymore, even after updating `api-list.yaml`.  I wasn't familiar enough with the project's workings to resolve it, and as my mandate was to just get the latest PubSub API I had to stop there.

Byron:
Great, please have a look, I believe PubSub should (also) have been republished.
The makefile was acting up for me as well, but I seem to have fixed so next time you try, it might actually work.
By now I am in 'brutal maintenance' mode, some kind of 'just get it to work' without necessarily understanding all the consequences of that. Let's hope it will all be fine :D.

joell:
I pulled the new PubSub crate and everything works smoothly so far.  Thanks, @Byron!

PR #262, opened 2021-02-04
Update to yup-oauth 5.0, newest hyper, rustls etc.

djrodgerspryor:
I switched to a new M1 mac and discovered the same problem with the old version of `ring` as you did @Byron.

I've updated the dependencies, and converted the APIs to use tokio 1 futures. I've tested the `sheets4` library (which is the only one I actually use) and it's working so far. I've built a couple of other libraries, but I haven't tested them much, and I certainly haven't even looked at many of them: I'd appreciate help from anyone who has a use case for them in testing them.

The one remaining problem is that the CLI crates are broken because `yup-hyper-mock` still depends on old crates. @Byron I'm not sure what the best approach here is: maybe we can remove it?

The code-generator changes have been broken-out into different commits from the changes to the generated code to make them easier to read.

### DIFF
--- a/src/mako/api/api.rs.mako
+++ b/src/mako/api/api.rs.mako
@@ -129,7 +129,7 @@ ${rbuild.new(resource, c)}
 // CallBuilders   ###
 // #################
 
-% for resource, methods in c.rta_map.iteritems():
+% for resource, methods in c.rta_map.items():
 % for method in methods:
 ${mbuild.new(resource, method, c)}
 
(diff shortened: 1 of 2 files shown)


Byron:
Sorry! It looks like the rewriting of history to remove the regenerated APIs from the PR has automatically closed it.
Could you reopen it with your latest master so we can get it to work and merged asap :)?

Thanks a lot for your help, my own efforts in #263 have stalled, but I am keen to get this to work and republish all crates.

djrodgerspryor:
Done: https://github.com/Byron/google-apis-rs/pull/264

PR #264, opened 2021-02-05
Update to yup-oauth 5.0, newest hyper, rustls etc. 

djrodgerspryor:
Here's a new PR with the generated files removed

PR #265, opened 2021-04-02
Bug fix: It was not possible to use FileGetCall to download binary files

dvaerum:
@Byron here is a pull request for the fix that we talked about in https://github.com/Byron/google-apis-rs/pull/264#issuecomment-811995650

I had originally made it as a pull request here:
https://github.com/djrodgerspryor/google-apis-rs/pull/1
but it didn't get merged in time, so now I will just make the merge request directly to you 😃

### DIFF
--- a/src/mako/api/lib/mbuild.mako
+++ b/src/mako/api/lib/mbuild.mako
@@ -811,25 +811,15 @@ else {
                     return Err(client::Error::HttpError(err))
                 }
                 Ok(mut res) => {
-                    let (res_parts, res_body) = res.into_parts();
-                    let res_body_string: String = String::from_utf8(
-                        hyper::body::to_bytes(res_body)
-                            .await
-                            .unwrap()
-                            .into_iter()
-                            .collect(),
-                    )
-                    .unwrap();
-                    let reconstructed_result =
-                        hyper::Response::from_parts(res_parts, res_body_string.clone().into());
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot, that's so much better. I also like the removal of the 'unwrap()' by allowing lossy conversion into an UTF-8 string in case of type deserialization.

I will make sure that the `drive3` API gets released with the fix one way or another - right now I have to publish in batches until they throttle me. Already reached out to crates.io to get my limit raised for this purpose.

In the meantime I thought maybe you are interested to [help with this small improvement](https://github.com/Byron/google-apis-rs/projects/1#card-58242291) which should help making everything just a little more efficient and remove unnecessary complexity. 5 years ago was quite a different time for `serde` spawning creative 'hacks' like these 😅.

Byron:
[Drive3](https://crates.io/crates/google-drive3/2.0.1+20210322) with the patch just arrived on crates.io. Thanks for all your help, google-APIs are back on track now!

dvaerum:
> In the meantime I thought maybe you are interested to [help with this small improvement](https://github.com/Byron/google-apis-rs/projects/1#card-58242291) which should help making everything just a little more efficient and remove unnecessary complexity. 5 years ago was quite a different time for `serde` spawning creative 'hacks' like these 😅.

@Byron I will give it a try

Issue #266, opened 2021-04-05
Large responses are truncated.


harababurel:
I've encountered this bug while trying to integrate `google-drive3 v2.0.1+20210322` on [harababurel/gcsf](https://github.com/harababurel/gcsf).

I'm creating a `FileListCall` to retrieve many files from Drive. The response I get seems to be truncated to exactly 16384 bytes. [Here](https://gist.github.com/harababurel/6d34e8f3f5927dfb3e254f84023467fa) are some logs showing the truncation.

I've pinned the root cause to `client::get_body_as_string` which only returns the first chunk in the hyper response:

https://github.com/Byron/google-apis-rs/blob/a3e2835a89876cd6e20a41ecce92a8a4a65cf8cc/gen/drive3/src/client.rs#L812-L816

[`hyper::body::aggregate`](https://docs.rs/hyper/0.14.5/hyper/body/fn.aggregate.html) returns an `impl Buf` which also provides [`remaining()`](https://docs.rs/hyper/0.14.5/hyper/body/trait.Buf.html#tymethod.remaining). This can probably be used in conjunction with `chunk()` to parse the entire buffer.

I'll try to implement the logic and send a PR some time this week.

Byron:
Thanks for this wonderful issue and analysis on how to fix it! I thought it's interesting that I still remember reading this `.chunk()` part thinking: "Aaaah, first you aggregate the body bytes which can be multiple chunks into a single chunk, makes sense" without reading docs or further questioning it, and being full aware that it would be an error prone API to have something aggregate `Bytes` into `Bytes` which then magically appear to have one chunk only.

Writing this out hopefully helps to me to listen to these glimpses of intuition next time and catch a bug, or two :D.

dvaerum:
@harababurel Thanks for fixing and explaining my mistake, I had problems figuring out which one to use `aggregate` or `to_bytes` and I clearly misunderstood the documentation :sweat_smile:

PR #267, opened 2021-04-05
Fix truncated responses + clean up (cargo clippy + cargo fmt)

harababurel:
See issue https://github.com/Byron/google-apis-rs/issues/266.


Byron:
All this looks great, thanks a lot. I will soon publish patch releases for `drive3` and `youtube3`.

PR #280, opened 2021-04-21
Bug-fix for the upload error: Failed at uploading 'client requires absolute-form URIs

dvaerum:
I found a bug in the code preventing be from uploading to Google Drive, the fix was pretty simple, one just need to add the `url` to the request, from some reason that was missing.

I assume that this bug also exist for all the other APIs there has an upload feature, because the code is the same for all of them.

**Bonus/Note:**
- Bumped the python lib used to generate everything to the newest version, I was unable to install the old ones any more. It seems to have caused by a deprecation of a library.
- Found some unused code, which I removed  
  `let url = upload_url.as_ref().and_then(|s| Some(url::Url::parse(s).unwrap())).unwrap();`

### DIFF
--- a/etc/bin/mako-render
+++ b/etc/bin/mako-render
@@ -233,7 +233,7 @@ def load_data(datafiles):
             if not imported_yaml:
                 import yaml
                 imported_yaml = True
-            data = yaml.load_all(open(filename, 'r'))
+            data = yaml.load_all(open(filename, 'r'), Loader=yaml.loader.FullLoader)
             data = list(data)[0]
         else:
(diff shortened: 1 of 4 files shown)


Byron:
Thanks a lot for the fixes, much appreciated :)!

I have republished the drive3 API at v2.0.5 for your convenience.

dvaerum:
> Thanks a lot for the fixes, much appreciated :)!

You are welcome :D

> I have republished the drive3 API at v2.0.5 for your convenience.

Thanks :tada:

PR #292, opened 2021-08-03
Correct multipart line endings

### LINKED ISSUE #291: MultiPartReader produces invalid multipart bodies
https://github.com/Byron/google-apis-rs/blob/main/gen/storage1/src/client.rs#L452

This write terminates the part headers with a single `\r\n` where a double `\r\n\r\n` is required. Consequentially every simple object insert call fails with status 400 and response body `Malformed multipart body.`.

The reader also incorrectly adds an `\r\n` at the beginning and omits the one at the end of the last line, but this doesn't seem to bother the Google server.

In an unrelated issue: the error gets consumed from the response stream but does not get parsed since it is not JSON. A `client::Error::Failure(response)` is returned, but the `response.body` can no longer be read and the original server error message is lost.

Recmo:
Fixes #291

### DIFF
--- a/gen/storage1/src/client.rs
+++ b/gen/storage1/src/client.rs
@@ -1,5 +1,3 @@
-// COPY OF 'src/rust/api/client.rs'
-// DO NOT EDIT
 use std::error;
 use std::fmt::{self, Display};
 use std::io::{self, Cursor, Read, Seek, SeekFrom, Write};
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot, much appreciated. I will see if I can republish storage1 today.

PR #299, opened 2021-09-29
Expose yup_oauth2 crate from inside the crates

### LINKED ISSUE #298: Expose oauth2 crate
Hello, and thank you for this crate !

The yup_oauth2 crate is required for any project that wishes to call ```DriveHub::new```. This could lead to some issues when new versions of yup_oauth2 are released. For example, version 6.0.0 of yup_oauth2 has been released, and it is impossible use ```yup_oauth2 = "*"``` and ```google_drive3 = "*"``` in the same Cargo.toml.

If the yup_oauth2 is exposed with a ```pub use yup_oauth2```, it removes the need having a dependency on yup_oauth2 in the first place, and solves this issue.

DeboMac:
Fix #298

### DIFF
--- a/src/mako/api/lib.rs.mako
+++ b/src/mako/api/lib.rs.mako
@@ -56,3 +56,5 @@ pub mod client;
 // Re-export the hub type and some basic client structs
 pub use api::${hub_type};
 pub use client::{Result, Error, Delegate};
+// Re-export the yup_oauth2 crate, that is required to call some methods of the hub and the client
+pub use oauth2;
--- a/src/mako/api/lib.rs.mako
+++ b/src/mako/api/lib.rs.mako
@@ -46,6 +46,8 @@ extern crate serde_derive;
 extern crate hyper;
 extern crate serde;
 extern crate serde_json;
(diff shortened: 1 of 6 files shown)


Byron (changes requested):
Thanks a lot for the fix!

One tiny request though, could you please not regenerate everything, so only the changes to the src/ folder are present in the PR.

That way we avoid this insanity 😅:
Screenshot 2021-09-30 at 10 18 11

DeboMac:
Yep, it makes more sense! 
Here is the PR with only source files modified

Byron:
This line should easily reproduce the issue locally for a fix:

```
make drive3-cli-cargo ARGS=check
```

Note that it also builds the CLI which might not need to declare yup-oauth2 in its manifest either.

Byron:
Thanks a lot, that works! Please let me know if there is any crate in particular that you'd like to see republished and it shall be done.

DeboMac:
Thank you for being so responsive ! I'd love to see google-drive3 published, if it is doable !

PR #301, opened 2021-10-21
support schema RECORD and REPEATED

tmokazaki:
If schema type has `RECORD` (and/or mode is `REPEATED`), query seems to be failed. So I tried to add TableCell additional types to decode these schema. I'm not sure, the change here is appropriated.

Please let me know if it is ok or not. Thank you in advance.


Byron:
Thanks for your contribution. Unfortunately it's in a file that is entirely generated and thus does not allow edits. The generator doesn't support such change either, so making this work here is probably difficult.

[`cargo-patch`](https://lib.rs/crates/cargo-patch) might be usable in this case to patch the package locally, or to have a manually maintained vendored version of the crate.

PR #319, opened 2022-03-09
bigquery: add non-uploading job.insert rpc

eiz:
BQ job.insert supports metadata-only job creation. I needed this for something.

not sure if I'm intended to include the gen'd code in the pr or if there's other tasks that need to be run.

(sorry, clicked the wrong branch...)


Byron:
Thanks a lot for the improvement, I am glad these overrides could be useful once again.

Rebuilding a single API is quite helpful, too, thanks for that.

The new bigquery release 3.0.1 contains the change.

PR #340, opened 2022-06-03
Fix a few minor things breaking GHA

DavidS:
This change is pulled from https://github.com/Byron/google-apis-rs/blob/c5ff239961df59cae23a7dd609d69b629f31e7fa/src/mako/cli/main.rs.mako#L39 to address the [GHA failures](https://github.com/Byron/google-apis-rs/runs/6718816121?check_suite_focus=true#step:3:532).


Byron:
Amazing! It looks like you single-handedly fixed CI while fixing the build issue! Thanks so much!
Please let me know if you need anything to be re-published.

PR #341, opened 2022-06-28
Update README.md

humb1t:
`HttpsConnector` has no methods described in example, changed to `HttpsConnectorBuilder` which has.


Byron:
Thanks a lot, much appreciated!

PR #343, opened 2022-07-12
Update outdated docs

lamafab:
The current docs shows a library interaction that is no longer valid. You now need to use `HttpsConnectorBuilder` instead of `HttpsConnector`.


lamafab:
I guess this should also be added to the create root documentation.

Byron (changes requested):
Thanks a lot!

> I guess this should also be added to the create root documentation.

Indeed, this is where it has to be added, it's in one of the `src/*.mako` files. Please refrain from regenerating the generated files, otherwise the PR becomes illegible 😅. 

Thank you

Byron:
Closing due to inactivity.

Issue #347, opened 2022-07-21
Panic on http2 timeout


JohnDimas:
hyper crate allows to set http2 timeouts within google_drive3 as follow : 
```rust
let hub = DriveHub::new(
    hyper::Client::builder()
        .http2_keep_alive_interval(Duration::from_secs(60))
        .http2_keep_alive_timeout(Duration::from_secs(60))
        .build(...),
    auth,
);
```

But if the connection got broken (for exemple during an upload), google_drive3 could panic at this unwrap()
https://github.com/Byron/google-apis-rs/blob/0a8b3ca4e465a5fc9a099682d4b074255912c204/gen/drive3/src/client.rs#L720-L724

with
```bash
[DEBUG hyper::proto::h2::client] connection keep-alive timed out
panicked at 'called `Result::unwrap()` on an `Err` value: hyper::Error(Body, Error { kind: Io(Kind(BrokenPipe)) })', /home/...google-drive3-4.0.0+.../src/client.rs:722:30
```

Solution: return an error instead of unwrap() at client.rs:722

PR #348, opened 2022-07-21
Resolves 'Panic on http2 timeout '

JohnDimas:
Fixes https://github.com/Byron/google-apis-rs/issues/347

### DIFF
--- a/gen/drive3/README.md
+++ b/gen/drive3/README.md
@@ -137,7 +137,7 @@ let auth = oauth2::InstalledFlowAuthenticator::builder(
         secret,
         oauth2::InstalledFlowReturnMethod::HTTPRedirect,
     ).build().await.unwrap();
-let mut hub = DriveHub::new(hyper::Client::builder().build(hyper_rustls::HttpsConnector::with_native_roots().https_or_http().enable_http1().enable_http2().build()), auth);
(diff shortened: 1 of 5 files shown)


Byron (changes requested):
Thanks a lot for tackling this!

Could you please translate the changes in `gen/` to the `src/` directory which contains the template source files that will apply to all generated files under `gen/`. Any changes to `gen/` will be overwritten.

Running `make` will provide an overview of possible targets, and in this case it's possible to test changes in the templates by running cargo on the generated output like so `make drive3-cli-cargo ARGS=check`.

Thanks

JohnDimas:
The only thing I did was to modify `src/rust/api/client.rs` and run `make regen-apis`. I did not modify any file in gen/ by myself :+1: 

Should I regen drive3-api by an other way ?
I did run `make drive3-cli-cargo ARGS=check` but everything seems fine

Byron:
Thanks for the clarification and for your help - it was confusing for me to see more changed files in `gen/` than in `src/`. I resolved this by undoing the changes in `gen/` first to regenerate them myself with similar results. I didn't look into why that is either 😅.

That said, you will find [drive 3 v4.0.1 on crates.io](https://crates.io/crates/google-drive3).

JohnDimas:
Thank you :rocket:

PR #353, opened 2022-08-20
Python fixes: Initial setup, type checking & tests

kylegentle:
To prepare for further refactoring of the Python codebase, this PR:

* Sets up basic Python type-checking capability with pyright, which was chosen over mypy or pyre for its support of recursive type definitions. Also define a new Make target for this.

* Renames `src/mako` to `src/generator`, and nests the mako templates inside a `src/generator/templates` dir. This isolates most regular Python code into the `src/generator/lib` dir.

* Updates `lib.util.new_context` to be compatible with regular `Dict` objects, as an incremental step from DictObject towards a typed dataclass implementation.

* Adds some tests for `lib.util.new_context`.

* Removes some dynamic imports & legacy libs in `etc/bin/mako-render`.


Byron:
Thanks so much for digging through all that cruft and making it better! I am most grateful.
Also I am looking forward to the future PRs you have planned.

Thank you!

PR #364, opened 2022-09-23
 Use generic Authy trait in place of yup_oauth2

philippeitis:
Note: New PR without generated API files.

I decided to give the "decouple" approach a try, as it fits my needs better.

I took the approach of creating an `Authenticator` trait which allows tokens to be generated asynchronously (as approached to the delegate, which requires sync). I believe the crates should still work as before for anyone not using a custom `Delegate` trait, as `Hub::new` accepts an instance of `Authenticator`, which is implemented for `yup_oauth2::authenticator::Authenticator`.

`yup-oauth2` is no longer a required dependency, and can be disabled if a user should choose to do so by using `default-features = false`. 

The only breaking change introduced is `err` parameter from the `Delegate::token` method. Given that the `Authenticator` implementation can handle any `yup-oauth2` errors internally, it doesn't seem necessary for `Delegate::token` to also handle this.

There's a default implementation of `Authenticator` for `String` (user-provided oauth tokens) and `()` (no token at all - this seems like a good idea for APIs which do not use need oauth2)

### DIFF
--- a/src/generator/templates/api/lib/mbuild.mako
+++ b/src/generator/templates/api/lib/mbuild.mako
@@ -707,13 +707,24 @@ else {
         loop {
             % if default_scope:
             let token = match ${auth_call}.get_token(&self.${api.properties.scopes}.keys().map(String::as_str).collect::<Vec<_>>()[..]).await {
-                Some(token) => token.clone(),
-                None => {
-                    match dlg.token() {
+                // TODO: remove Ok / Err branches
+                Ok(Some(token)) => token.clone(),
+                Ok(None) => {
+                    let err = oauth2::OtherError("unknown error occurred while generating oauth2 token".into());
+                    match dlg.token(&err) {
                         Some(token) => token,
                         None => {
                             ${delegate_finish}(false);
-                            return Err(client::Error::MissingToken)
+                            return Err(client::Error::MissingToken(err))
+                        }
+                    }
+                }
+                Err(err) => {
+                    match dlg.token(&err) {
+                        Some(token) => token,
+                        None => {
+                            ${delegate_finish}(false);
+                            return Err(client::Error::MissingToken(err))
                         }
                     }
                 }
--- a/src/generator/templates/api/lib/mbuild.mako
+++ b/src/generator/templates/api/lib/mbuild.mako
@@ -387,7 +387,7 @@ match result {
          Error::HttpError(_)
         |Error::Io(_)
         |Error::MissingAPIKey
-        |Error::MissingToken
+        |Error::MissingToken(_)
         |Error::Cancelled
         |Error::UploadSizeLimitExceeded(_, _)
         |Error::Failure(_)
@@ -710,7 +710,7 @@ else {
                 // TODO: remove Ok / Err branches
                 Ok(Some(token)) => token.clone(),
                 Ok(None) => {
-                    let err = oauth2::OtherError("unknown error occurred while generating oauth2 token".into());
+                    let err = oauth2::Error::OtherError(anyhow::Error::msg("unknown error occurred while generating oauth2 token"));
                     match dlg.token(&err) {
                         Some(token) => token,
                         None => {
--- a/src/generator/templates/api/lib/mbuild.mako
+++ b/src/generator/templates/api/lib/mbuild.mako
@@ -387,7 +387,7 @@ match result {
          Error::HttpError(_)
         |Error::Io(_)
         |Error::MissingAPIKey
-        |Error::MissingToken(_)
+        |Error::MissingToken
         |Error::Cancelled
         |Error::UploadSizeLimitExceeded(_, _)
         |Error::Failure(_)
@@ -706,14 +706,14 @@ else {
 
         loop {
             % if default_scope:
-            let token = match ${auth_call}.token(&self.${api.properties.scopes}.keys().collect::<Vec<_>>()[..]).await {
-                Ok(token) => token.clone(),
-                Err(err) => {
-                    match  dlg.token(&err) {
+            let token = match ${auth_call}.get_token(&self.${api.properties.scopes}.keys().map(String::as_str).collect::<Vec<_>>()[..]).await {
+                Some(token) => token.clone(),
+                None => {
+                    match dlg.token() {
                         Some(token) => token,
                         None => {
                             ${delegate_finish}(false);
-                            return Err(client::Error::MissingToken(err))
+                            return Err(client::Error::MissingToken)
                         }
                     }
                 }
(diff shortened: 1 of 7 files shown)


src/rust/api/client.rs
```diff
@@ -785,3 +786,75 @@ pub async fn get_body_as_string(res_body: &mut hyper::Body) -> String {
...
+        self.clone_box()
+    }
+}
+
+impl Authy for String {
+    fn get_token<'a>(&'a self, _scopes: &'a [&str]) -> Pin<Box<dyn Future<Output=Option<String>> + 'a>> {
+        async fn ident(s: &str) -> Option<String> {
+            Some(s.to_string())
+        }
+        Box::pin(ident(self))
```
Byron, line 824: I think `async move {self.clone()}` should do the trick.

src/rust/api/client.rs
```diff
@@ -785,3 +786,75 @@ pub async fn get_body_as_string(res_body: &mut hyper::Body) -> String {
...
+    use tokio::io::{AsyncRead, AsyncWrite};
+    use http::Uri;
+    use hyper::client::connect::Connection;
+
+    async fn helper<S>(auth: &Authenticator<S>, scopes: &[&str]) -> Option<String> where
+        S: Service<Uri> + Clone + Send + Sync + 'static,
+        S::Response: Connection + AsyncRead + AsyncWrite + Send + Unpin + 'static,
+        S::Future: Send + Unpin + 'static,
+        S::Error: Into<Box<dyn std::error::Error + Send + Sync>> {
+        auth.token(scopes).await.ok().map(|t| t.as_str().to_string())
```
Byron, line 848: `t.as_str().to_string()` should be `t.as_str().to_owned()`.

src/generator/templates/api/lib/mbuild.mako
```diff
@@ -706,14 +706,14 @@ else {
 
         loop {
             % if default_scope:
-            let token = match ${auth_call}.token(&self.${api.properties.scopes}.keys().collect::<Vec<_>>()[..]).await {
-                Ok(token) => token.clone(),
-                Err(err) => {
-                    match  dlg.token(&err) {
+            let token = match ${auth_call}.get_token(&self.${api.properties.scopes}.keys().map(String::as_str).collect::<Vec<_>>()[..]).await {
+                Some(token) => token.clone(),
```
Byron, line 710: As nice as it would be to adapt the return value to the new trait, I would feel much more comfortable to _not_ make this a breaking change.

This means `get_token()` has to return `Result<Option<String>, yup_oauth2::Error` which should then be handled as before. Those custom implementations simply don't use it, and those who have errors can find one to use in `yup_oauth2::Error`. This also means that `yup-oauth2` is still a required dependency.

It would be fair to mark this as 'TODO' since in the next major release it should be cleaned up to be like it is in this PR.

Byron (changes requested):
Thanks a lot, this looks great and I see how this would be working.

I have a few minor notes and a major (but easy to do) change to not make it a breaking change just yet.

When trying to run the generator locally (this review was just via the browser), I ran into a problem with python too, maybe even the same as you? When installing `Markdown` it errors at `AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'getiterator'`, probably because the python version used is too new and they have breaking changes between minor versions, or maybe even patches.

How did you workaround it? Install an older python version? (_I am asking because I didn't want to spend any time on dealing with python, I am done with it 😅_)

Thanks for your help

philippeitis:
I've included my fix for the python build error (updating mkdoc to 1.3.1 - that being said, I don't know what mkdoc does in this case, so I'm not sure if this would cause any issues).

I've removed the breaking changes, with TODO notes marking all the necessary spots to change back when the next major version comes along. 

I ultimately had two major questions:

The first question is if there was any particular reason `Delegate` doesn't handle authentication directly - ultimately, if breaking changes are to be made, I think it'd be ideal if `Delegate` handled everything and `GetToken` instances could either be converted into a `Delegate` builder, or implement the `Delegate` trait. In this case, I'd first change `String` to `AccessToken`, to make the `Delegate` conversion obvious. 

The second question is if it'd be possible to split out the shared functionality (eg. client.rs)  in each crate into a common backend, and re-export as necessary. This would allow sharing delegates amongst multiple hubs (eg. I build an application which downloads files from one service, transforms them, and uploads them to another), which is not currently possible as each `Delegate` trait is unique.

Byron:
> I've included my fix for the python build error (updating mkdoc to 1.3.1 - that being said, I don't know what mkdoc does in this case, so I'm not sure if this would cause any issues).

Thanks so much! I can run it fine locally but CI for some reason has problems. It's strange as versions are pinned and I'd expect reproducible results. Then again, it's python, so all bets are off.

Thanks for all the updates, they look good at a glance but I will take a more thorough look later before merging.

> The first question is if there was any particular reason Delegate doesn't handle authentication directly - ultimately, if breaking changes are to be made, I think it'd be ideal if Delegate handled everything and GetToken instances could either be converted into a Delegate builder, or implement the Delegate trait. In this case, I'd first change String to AccessToken, to make the Delegate conversion obvious.

I am absolutely open to that, even though I feel putting auth into the delegate directly it fattening it even more. `GetToken` is a neat abstraction. The true reasoning for the status quo is definitely lost in the tides of time, but I'd guess that `yup-oauth2` was separate enough to warrant it remaining separate. Back in the days, the crate was still owned by me so that was all control I needed.

> The second question is if it'd be possible to split out the shared functionality (eg. client.rs) in each crate into a common backend, and re-export as necessary. This would allow sharing delegates amongst multiple hubs (eg. I build an application which downloads files from one service, transforms them, and uploads them to another), which is not currently possible as each Delegate trait is unique.

This is definitely an oversight fixing this should be non-breaking, too. If you'd like this can be addressed in a second PR. I am thinking to put the code into a separate crate and re-export that as usual via the API crates.

Thanks again for your quick response and your work on this project, it's much appreciated and contributions are the reason these crates are still relevant. 

PS: CI still fails with `mkdocs` seeing an invalid `nav` structure. It can be reproduced with `make gen/youtube3-cli/build_html/index.html`, and then `cd gen/youtube3-cli;  <root>/.pyenv-<platform>/bin/mkdocs build --clean` in case you could give it a shot. The generator code for the navigation [is here](https://github.com/philippeitis/google-apis-rs/blob/fedf48445febb012d50286c9d73808cbf4cc636e/src/generator/templates/cli/mkdocs.yml.mako#L17:L17).

Byron:
That's fantastic work, thank you!

If this is enough, please let me know and I can republish crates of your interest. Alternatively, we can do that once `client.rs` has been moved into its own crate. Please let me know.

PR #365, opened 2022-09-25
Split out common api client code into separate crate

philippeitis:
I've separated out the API client code into a separate crate.

Due to the necessary changes to dependencies, I've also removed some spurious `extern crate` usages, and replace `serde-derive` with `serde`'s `derive` feature (identical functionality). 

Given that `extern crate` isn't necessary as of the 2018 edition, I think removing this should not cause any issues.

I'm not entirely sure how to remove the copied `client.rs` file (which seems to be generated by `deps.mako`, but I'm not entirely sure how this file works).


Byron (approved):
Thanks a million for pushing `google-apis-rs` this far!

I made a few adjustments, most notably the name change of `google-api-client` to `google-apis-common`. Having taken a quick look at the code that copies `client.rs` around, I decided it's easiest to keep it as is but neuter the `api/client.rs` file and explain why it still exists, despite being technically useless.

In a moment I will merge this PR and publish the new crate. If you let me know which crates you'd like to see published, I am happy to do that as well right after.

PR #366, opened 2022-09-28
Common CLI crate

philippeitis:
Moving client.rs to a common crate has a few benefits - the primary of which is that it reduces number of files (as well as the number of changes) in the repository. However, additional (but small) benefits include: easier updates to client functionality for all CLI crates, slightly faster compile times, and a slightly simplified `deps.mako` file


Byron:
Thanks a million, great work!

I have published drive3 API and CLI as v4.0.2 with all the latest improvements.

PR #369, opened 2022-09-29
Add missing trait bounds for proper async to GetToken

philippeitis:
I apparently missed a few necessary trait bounds while implementing `GetToken`, which I discovered while using the crate in my project. I've added the necessary bound plus a test to ensure that it remains correct in the future.

Unfortunately, I think this might necessitate yanking the previous versions of google-apis-common (as they will likely fail to compile).


Byron:
Thanks a lot! I am glad I didn't republish all crates just yet as well :). 

This fix should resolve #372 as well, and make obsolete #373.

Byron:
google-drive3 was republished as patch release, containing the fix.

PR #371, opened 2022-09-30
Refactor client functions

philippeitis:
I'm not sure whether these changes should be included in the previous PR - but given that it's non-essential, I've decided to keep them separate. 

The first change is refactoring `remove_json_null_values` to run in linear time (the `arr.remove()` call takes `O(n)` time and may be  called `n` times, which could accidentally lead to quadratic run times) - this incidentally makes the code a lot cleaner.

The second change is using `tokio::time::sleep` in place of `std::thread::sleep` in the upload functions (for both `google-apis-common` and `api.rs`). `std::thread::sleep` is blocking per https://doc.rust-lang.org/stable/std/thread/fn.sleep.html, and should not be used in async functions.

### DIFF
--- a/google-apis-common/src/lib.rs
+++ b/google-apis-common/src/lib.rs
@@ -5,7 +5,6 @@ use std::future::Future;
 use std::io::{self, Cursor, Read, Seek, SeekFrom, Write};
 use std::pin::Pin;
 use std::str::FromStr;
-use std::thread::sleep;
 use std::time::Duration;
 
 use itertools::Itertools;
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a million for these fixes! Especially the `thread::sleep` one is a huge oversight that could make these crates unusable in a server context, a lingering time-bomb if you will.

Too bad I forgot about this PR when publishing previously, so maybe in future it's preferred to keep small fixes with unrelated PRs as my memory can't be trusted 😅.

Issue #377, opened 2022-10-05
Parse more values to appropriate types


philippeitis:
Certain values don't currently use the appropriate types as described by the schema - it seems that due to JSON limitations, Google might use strings to encode numbers - a few examples below, taken from the `drive-api.json` file:

```json
"version": {
    "description": "A monotonically increasing version number for the file. This reflects every change made to the file on the server, even those not visible to the user.",
    "format": "int64",
    "type": "string"
}
```

```json
"durationMillis": {
    "description": "The duration of the video in milliseconds.",
    "format": "int64",
    "type": "string"
}
```

```json
"size": {
    "description": "The size of the revision's content in bytes. This is only applicable to files with binary content in Drive.",
    "format": "int64",
    "type": "string"
}
```

This results in me having to write code like this:

```let size = u64::from_str(&file.size.unwrap()).unwrap();```

Even though size should really be parsed by the API.

This seems to be due to this map: 
https://github.com/Byron/google-apis-rs/blob/93c8601fdca62f92f164957d9e86220f4f1b10f7/src/generator/lib/util.py#L22

And this block of code:
https://github.com/Byron/google-apis-rs/blob/93c8601fdca62f92f164957d9e86220f4f1b10f7/src/generator/lib/util.py#L380

I think the fix would be to default to using the format specifier first, falling back to the type specifier if the format is not present / supported. 

However, this would require a breaking change, so I want to ask about how this should be handled before making any changes.

Byron:
Thanks a lot for bringing this up! I remember having left the type-system to be just barely good enough to be acceptable for the use-case I had back then, and agree that improvements are desirable and possible.

> However, this would require a breaking change, so I want to ask about how this should be handled before making any changes.

I am absolutely open to that, it seems to be a worthy cause. Even though the version number was incremented to 4.0 not too long ago, the crates have not actually been republished. By going to 5.0 it would mean people have to deal with a couple of additional breaking changes, which I think is probably worth it for them if they consider the upgrade in the first place.

> It would also be nice to support "date-time" format values, but that would require pulling in additional dependencies, which may not be desirable. I think it would be nice to provide base64 decoding / encoding for "byte" format values, to ensure that those values are constructed correctly.

I definitely wouldn't mind, and rather throw CPU time at a problem than human time. If there is high demand, a feature toggle could be added later to avoid those newly added dependencies.

A PR with such an improvement is definitely welcome.

PR #379, opened 2022-10-07
Use schema format over type

philippeitis:
This is a draft PR which includes the basic fix, allowing more values to be deserialized correctly. 

To handle Google specific types, I've included a rudimentary implementation for base64 and `Duration` encode/decode.

The `Duration` struct can simply be imported where necessary, and implements serde directly.
However, the base64 encoding / decoding requires custom attributes to utilize the appropriate serde functions (as we'd want it to be `Vec<u8>`). 

This leaves 3 date-time types, and one fieldmask type. Unfortunately, it is somewhat difficult to track down the concrete type definitions, so I'm not entirely confident that the implementations thus far are correct.

However, I think this go package might be relevant: https://pkg.go.dev/google.golang.org/protobuf/types

(Incidentally, it would be nice to support protobuf APIs for supported endpoints - something I'd like to look at in the future)

PR for #377

### DIFF
--- a/google-apis-common/Cargo.toml
+++ b/google-apis-common/Cargo.toml
@@ -29,6 +29,3 @@ hyper = "^ 0.14"
 http = "^0.2"
 tokio = "^1.0"
 tower-service = "^0.3.1"
-
-[dev-dependencies]
-serde = { version = "^ 1.0", features = ["derive"] }
--- a/google-apis-common/Cargo.toml
+++ b/google-apis-common/Cargo.toml
@@ -20,6 +20,7 @@ mime = "^ 0.2.0"
 serde = { version = "^ 1.0", features = ["derive"] }
 base64 = "0.13.0"
 serde_json = "^ 1.0"
+chrono = { version = "0.4.22", features = ["serde"] }
 ## TODO: Make yup-oauth2 optional
 ## yup-oauth2 = { version = "^ 7.0", optional = true }
 yup-oauth2 = "^ 7.0"
--- a/google-apis-common/Cargo.toml
+++ b/google-apis-common/Cargo.toml
@@ -17,7 +17,8 @@ doctest = false
 
 [dependencies]
 mime = "^ 0.2.0"
-serde = "^ 1.0"
+serde = { version = "^ 1.0", features = ["derive"] }
+base64 = "0.13.0"
 serde_json = "^ 1.0"
 ## TODO: Make yup-oauth2 optional
 ## yup-oauth2 = { version = "^ 7.0", optional = true }
--- a/google-apis-common/Cargo.toml
+++ b/google-apis-common/Cargo.toml
@@ -18,9 +18,12 @@ doctest = false
 [dependencies]
 mime = "^ 0.2.0"
 serde = { version = "^ 1.0", features = ["derive"] }
-base64 = "0.13.0"
+serde_with = "2.0.1"
 serde_json = "^ 1.0"
+
+base64 = "0.13.0"
 chrono = { version = "0.4.22", features = ["serde"] }
+
 ## TODO: Make yup-oauth2 optional
 ## yup-oauth2 = { version = "^ 7.0", optional = true }
 yup-oauth2 = "^ 7.0"
--- a/google-apis-common/src/field_mask.rs
+++ b/google-apis-common/src/field_mask.rs
@@ -1,0 +1,121 @@
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+fn titlecase(source: &str, dest: &mut String) {
+    let mut underscore = false;
+    for c in source.chars() {
+        if c == '_' {
+            underscore = true;
+        } else if underscore {
+            dest.push(c.to_ascii_uppercase());
+            underscore = false;
+        } else {
+            dest.push(c);
+        }
+    }
+}
+
+fn snakecase(source: &str) -> String {
+    let mut dest = String::with_capacity(source.len() + 5);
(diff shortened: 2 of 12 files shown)


philippeitis:
I've identified that the JSON representation for `google-datetime` and `datetime` follows RFC3339, which `chrono` implements `serde` for. `date` is simply a truncated version of RFC3339, and also parsed by `chrono`.

I've managed to use `chrono::Duration` for `google-duration`, with custom `serde` implementations.

I can't find a source which states that `byte` is universally representated as url-safe base64.

The last type is `google-fieldmasks`, but this is a lower priority for me.

I also need to update the generated READMEs to use the correct constructors.

philippeitis:
That should be all of the previously unsupported types, with appropriate unit tests and example types. While I still want to do some testing on actual API calls, this PR should be mostly ready to go.

I've included some links / information about where I found type descriptions/definitions in `util.py`, but I'm happy to put these links somewhere else.

Byron (changes requested):
Thanks a lot for your incredible effort, they are, as always, much appreciated!

I glanced at the changes and they looked alright, while running `make cargo-api ARGS=check` in the background. It stopped at `accesscontextmanager1` with:

```
    Checking google-accesscontextmanager1 v4.0.4+20220301 (/Users/byron/dev/github.com/Byron/google-apis-rs/gen/accesscontextmanager1)
error[E0277]: the trait bound `FieldMask: Clone` is not satisfied
   --> src/api.rs:957:5
    |
950 | #[derive(Default, Clone, Debug, Serialize, Deserialize)]
    |                   ----- in this derive macro expansion
...
957 |     pub update_mask: Option<client::FieldMask>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `FieldMask`
    |
    = note: required because of the requirements on the impl of `Clone` for `std::option::Option<FieldMask>`
```

Can you reproduce this and maybe get it to run through on all APIs as well? Then I'd think a merge shouldn't be far.

> I've included some links / information about where I found type descriptions/definitions in util.py, but I'm happy to put these links somewhere else.

To me this location seems like a great starting point and it felt intuitive to read them there. I'd keep them there until there is demand for a change or a vastly better place is presenting itself.

Thanks a lot!

philippeitis:
I've fixed that issue, and have made the custom serde implementations sufficiently general to support any nested type (eg. instead of only supporting `Option<i64>`, it now supports `Option<Vec<i64>>` or `Option<HashMap<String, client::chrono::Duration>>`). This required me to add a dependency on `serde_with`, and use a `RustType` class to represent the generated Rust types.

However, I've run into parameters converting unsupported types to string - eg. `firestore` transaction IDs are represented as bytes. The main problem is that `Vec<u8>` can not be converted to string, and I'm not sure how to change the generator to stringify the parameters correctly.

```
error[E0599]: `Vec<u8>` doesn't implement `std::fmt::Display`
    --> src/api.rs:6571:47
     |
6571 |             params.push(("transaction", value.to_string()));
     |                                               ^^^^^^^^^ `Vec<u8>` cannot be formatted with the default formatter
     |
    ::: /home/philippe/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:400:1
     |
400  | pub struct Vec<T, #[unstable(feature = "allocator_api", issue = "32838")] A: Allocator = Global> {
     | ------------------------------------------------------------------------------------------------
     | |
     | doesn't satisfy `Vec<u8>: ToString`
     | doesn't satisfy `Vec<u8>: std::fmt::Display`
     |
     = note: the following trait bounds were not satisfied:
             `Vec<u8>: std::fmt::Display`
             which is required by `Vec<u8>: ToString`
             `[u8]: std::fmt::Display`
             which is required by `[u8]: ToString`
```

This is because transaction IDs are bytes,

philippeitis:
I've managed to get `make cargo-api ARGS=check` to compile everything successfully. A few things to note:

1. I've added a dependency on `serde_with` to support types with arbitrary (de)serialization needs (eg. `Option<HashMap<String, client::chrono::Duration>>`) - without `serde_with`, I'd have to either introduce a wrapper type for `Vec<u8>` and `client::chrono::Duration`, which would add friction to the API, or add additional modules to support every special type that occurs.
2. I've introduced a `rust_type.py` file: this is used in conjunction with `serde_with` to support all arbitrary types - the `RustType` class provides functionality for providing the appropriate `serde_with` type and indicating if special serialization support is necessary.
3. I've introduced code to use the correct `to_string` implementation for parameter types - this is necessary to support `Vec<u8>` as urlsafe-base64, and to use the correct string format for datetimes and durations.

Byron:
This is amazing work! I'd even call it 'magical' as I always have trouble wrapping my head about `serde` internals, yet you seem to bend it to your will!

Introducing `serde-width` seems like a good choice and is well worth the added dependency. By now you are the authority on the type-system in these crates, and I don't claim to understand it anymore remembering only that back in the days when I wrote it, I just 'hammered' it to work well enough and called it a day. Now it seems more sophisticated and capable than ever, many thanks for that!

If you would ever be open to a maintainer role in this project, I'd be happy to send you an invite.

Byron:
`drive3 v5.0` was released as a sample crate. My plan is to iterate on it or other samples until the code is proven to work well enough  and then release all crates as v5.0. As I am not using any of the crates myself I will rely on feedback to determine the 'all-in' release date. Thanks.

PR #381, opened 2022-10-13
Code quality

philippeitis:
This PR includes some smaller fixes to the internal implementations. My end-goal is to simplify the logic and move as much functionality as possible into reusable functions, in order to reduce the binary size of the underlying libraries and ensure correctness by allowing individual components to be unit tested.

This includes 4 commits, each of which are hopefully self-explanatory - I've decided to include them in one PR to keep things simple.

I think the `add_scope` function should be simplified / refactored somewhat, as it both doesn't match the documentation, and should take a simpler set of parameters. In particular, the documentation states that `None` values will clear the scopes, but they don't actually do this. This also seems like an overly complicated implementation, and I think the API should be changed as follows:

- add a `clear_scopes` function (or similar), which explicitly clear the parameters
- Change the input of `add_scope` to a simple `Into<String>` or `AsRef<str>` (no option wrapper) 

As for being a project maintainer, which you previously mentioned - I'd be happy to help, especially considering that I'd like to make further changes.

### DIFF
--- a/src/generator/lib/util.py
+++ b/src/generator/lib/util.py
@@ -1242,13 +1242,14 @@ def size_to_bytes(size):
 
 
 def string_impl(p):
+    """Returns a function which will convert instances of p to a string"""
     return {
-        "google-duration": "::client::serde::duration::to_string",
-        "byte": "::client::serde::urlsafe_base64::to_string",
-        "google-datetime": "::client::serde::datetime_to_string",
-        "date-time": "::client::serde::datetime_to_string",
-        "google-fieldmask": "(|x: &client::FieldMask| x.to_string())"
-    }.get(p.get("format"), "(|x: &dyn std::fmt::Display| x.to_string())")
+        "google-duration": lambda x: f"::client::serde::duration::to_string(&{x})",
+        "byte": lambda x: f"::client::serde::urlsafe_base64::to_string(&{x})",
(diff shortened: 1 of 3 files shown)


Byron:
> This PR includes some smaller fixes to the internal implementations. My end-goal is to simplify the logic and move as much functionality as possible into reusable functions, in order to reduce the binary size of the underlying libraries and ensure correctness by allowing individual components to be unit tested.

❤️

> I think the add_scope function should be simplified / refactored somewhat, as it both doesn't match the documentation, and should take a simpler set of parameters. In particular, the documentation states that None values will clear the scopes, but they don't actually do this. This also seems like an overly complicated implementation, and I think the API should be changed as follows:

> add a clear_scopes function (or similar), which explicitly clear the parameters
Change the input of add_scope to a simple Into<String> or AsRef<str> (no option wrapper)

Agreed! Let's do this on top of the current v5.0 (without another major version bump) as this version isn't distributed to crates.io yet (except for drive3).

> As for being a project maintainer, which you previously mentioned - I'd be happy to help, especially considering that I'd like to make further changes.

Great to hear! I have invited you.

I definitely appreciate seeing PRs for changes so I can stay in the loop, but also won't mind if mundane changes go directly in to `main` at your own discretion.

Besides that, I am continuously amazed by your contributions and thankful that you chose to move this project forward!

Issue #333, opened 2022-04-14
The abbility to add mutliple scopes for a Call


elpiel:
I've made this small Trait extension for adding multiple scopes at once to a specific *Call.
It really helps when you have a static defining your scopes per request.

Can this be added to the *Call structs for convenience in some way?

```
use gmail1::api::{Scope, UserMessageGetCall, UserMessageListCall};

/// Extension trait for all Gmail calls.
pub trait CallExt {
    fn add_scopes(self, scopes: &[Scope]) -> Self;
}

impl<'a> CallExt for UserMessageGetCall<'a> {
    fn add_scopes(mut self, scopes: &[Scope]) -> UserMessageGetCall<'a> {
        for scope in scopes {
            self = self.add_scope(scope);
        }

        self
    }
}
impl<'a> CallExt for UserMessageListCall<'a> {
    fn add_scopes(mut self, scopes: &[Scope]) -> UserMessageListCall<'a> {
        for scope in scopes {
            self = self.add_scope(scope);
        }

        self
    }
}
```

Byron:
This does indeed look useful to have, and a PR would be very welcome. If you consider submitting one, please don't regenerate the APIs as it makes the review difficult. Thanks.

PR #382, opened 2022-10-15
Refactor modification of scopes

philippeitis:
- `add_scope` is simplified to simply add scopes, and no longer takes an `Option<_>` item.
- Added `add_scopes` convenience function, which allows adding multiple scopes at once. 
- Added `clear_scopes` function to explicitly allow clearing method scopes.

This also includes small improvements to documentation - primarily by including links to referenced types / methods

Addresses #333

### DIFF
--- a/src/generator/lib/util.py
+++ b/src/generator/lib/util.py
@@ -24,35 +24,6 @@ CHRONO_PATH = "client::chrono"
 CHRONO_DATETIME = f"{CHRONO_PATH}::DateTime<{CHRONO_PATH}::offset::Utc>"
 CHRONO_DATE = f"{CHRONO_PATH}::NaiveDate"
 USE_FORMAT = 'use_format_field'
-TYPE_MAP = {
-    'boolean': 'bool',
-    'integer': USE_FORMAT,
-    'number': USE_FORMAT,
-    'uint32': 'u32',
-    'double': 'f64',
-    'float': 'f32',
-    'int32': 'i32',
-    'any': 'String',  # TODO: Figure out how to handle it. It's 'interface' in Go ...
-    'int64': 'i64',
-    'uint64': 'u64',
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a lot for this contribution, so much better!

Drive3 v5.0.1 was released as canary if you will, and the version prior to it was yanked. We can keep doing that with other breaking changes and when it seems to stabilize we can re-release all crates.

Speaking of, something I didn't remember doing this year was to update all API specifications, while possibly pulling in entirely new APIs. This would definitely be good to do before releasing all crates and it's something you could do if it interests you. Let's get there when we get there :).

philippeitis:
Sure, that sounds good. However, yanking crates seems like a bad way of going about beta testing. I think you can attach a suffix like "-beta.(beta-version)" to show that a particular crate version isn't ready for release:

Example of `clap`:
https://docs.rs/clap/4.0.0-rc.2/clap/index.html
This would cause less disruption and clearly signal to end-users that the current crate make undergo breaking changes.

Byron:
> However, yanking crates seems like a bad way of going about beta testing. I think you can attach a suffix like "-beta.(beta-version)" to show that a particular crate version isn't ready for release:

I agree, especially as having a direct dependency to a yanked crate would probably _force_ a user to upgrade (at least it's explicit) to a breaking change. Beta-versions are tricky in `cargo` unfortunately as it will happily upgrade between beta versions, beta-2 is just a patch to beta-1 for example, but I don't know what it does with 5.0.1-beta and 5.0.2-beta. My guess is it will just upgrade these automatically on the next `cargo update` as well. 

That said, and despite probably being unable to completely prevent breakage while `cargo` works that way, adding a `beta-X` suffix (or maybe even `alpha`) is explicit and should be preferred.

PR #383, opened 2022-10-17
Auth refactor

philippeitis:
Since we're moving on to the next API version, I'm revisiting this change as well.

PR includes:
- Documentation on how to use authentication, and how to implement custom authenticators as needed
- More flexible authentication error types
- Optional `yup-oauth2` (included by default)
- Optional token for APIs which optionally require token.

I'd also like to see about entirely removing the `yup-oauth2` dependency for (and `auth` argument in Hub constructor) for unauthenticated APIs

### DIFF
--- a/google-apis-common/src/auth.rs
+++ b/google-apis-common/src/auth.rs
@@ -1,0 +1,96 @@
+use std::future::Future;
+use std::pin::Pin;
+
+// TODO: Simplify this to Option<String>
+type TokenResult = Option<String>;
+
+pub trait GetToken: GetTokenClone + Send + Sync {
+    /// Called whenever there is the need for an oauth token after
+    /// the official authenticator implementation didn't provide one, for some reason.
+    /// If this method returns None as well, the underlying operation will fail
+    fn get_token<'a>(
+        &'a self,
+        _scopes: &'a [&str],
+    ) -> Pin<Box<dyn Future<Output = TokenResult> + Send + 'a>> {
+        Box::pin(async move { None })
+    }
+}
+
+pub trait GetTokenClone {
+    fn clone_box(&self) -> Box<dyn GetToken>;
+}
+
+impl<T> GetTokenClone for T
+where
+    T: 'static + GetToken + Clone,
+{
+    fn clone_box(&self) -> Box<dyn GetToken> {
+        Box::new(self.clone())
+    }
+}
+
+impl Clone for Box<dyn GetToken> {
+    fn clone(&self) -> Box<dyn GetToken> {
+        self.clone_box()
+    }
+}
+
+impl GetToken for String {
+    fn get_token<'a>(
+        &'a self,
+        _scopes: &'a [&str],
+    ) -> Pin<Box<dyn Future<Output = TokenResult> + Send + 'a>> {
+        Box::pin(async move { Some(self.clone()) })
+    }
+}
+
+/// In the event that the API endpoint does not require an oauth2 token, `NoToken` should be provided to the hub to avoid specifying an
+/// authenticator.
+#[derive(Default, Clone)]
+pub struct NoToken;
+
+impl GetToken for NoToken {}
+
+// TODO: Make this optional
+// #[cfg(feature = "yup-oauth2")]
+mod yup_oauth2_impl {
+    use core::future::Future;
+    use core::pin::Pin;
+
+    use super::{GetToken, TokenResult};
+
+    use http::Uri;
+    use hyper::client::connect::Connection;
+    use tokio::io::{AsyncRead, AsyncWrite};
+    use tower_service::Service;
+    use yup_oauth2::authenticator::Authenticator;
+
+    impl<S> GetToken for Authenticator<S>
+    where
+        S: Service<Uri> + Clone + Send + Sync + 'static,
+        S::Response: Connection + AsyncRead + AsyncWrite + Send + Unpin + 'static,
+        S::Future: Send + Unpin + 'static,
+        S::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    {
+        fn get_token<'a>(
+            &'a self,
+            scopes: &'a [&str],
+        ) -> Pin<Box<dyn Future<Output = TokenResult> + Send + 'a>> {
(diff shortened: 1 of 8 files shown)


google-apis-common/src/lib.rs
```diff
@@ -777,86 +769,14 @@ pub async fn get_body_as_string(res_body: &mut hyper::Body) -> String {
     res_body_string.to_string()
 }
 
-// TODO: Simplify this to Option<String>
-type TokenResult = std::result::Result<Option<String>, oauth2::Error>;
```
Byron, line 781: Something I'd like to see in the docs is how to obtain the reason for authentication failure. Especially when using `yup-oauth2`, it's easy to botch auth and knowing what failed exactly helps to resolve such issues.

My suggestion is to either keep this as `Result<Option<` but turn the error into `Box<dyn std::error::Error + Send + Sync + 'static>`, or to adjust the `GetToken` trait to query errors for the past `get_token()` call, which would then be used by the generated code to report errors.

To me it seems that leaving this as `Result<Option` is much more straightforward to not degenerate information. From my experience that's the way to go and the best way to indicate that `something can fail, and it might not exist, and checking for existence might fail itself` (hence, `Result<Option<` instead of `Option<Result<`).

Byron (changes requested):
This is great work and I love the direction this is going!

The reason it's set to `Request changes` is merely to discuss a few concerns mentioned in the comments before merging. Thank you.

google-apis-common/src/auth.rs
```diff
@@ -0,0 +1,163 @@
...
+//! # Usage
+//! [`GetToken`] instances are designed to be used with the Hub constructor provided by the
+//! generated APIs.
+//!
+//! If you intend to use the API libraries on client devices,
+//! [`Authenticator`] supports a variety of client-side authentication methods,
+//! and should be used to provide authentication.
+//!
+//! If you intend to use the API libraries server-side, with server-side client authentication,
+//! use the [`oauth2`] crate and convert the resulting [`AccessToken`] to [`String`].
```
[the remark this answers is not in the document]
philippeitis: I think you're referring specifically to service accounts - essentially, an account with permission to access a specific set of services. I believe service accounts can be given permission to access any Google API, but are intended mainly for developer / backend use.
 
However, I'm specifically referring to accessing APIs like Google Drive on the behalf of an unknown end user from a remote server (server gets token, accesses Google Drive). This use-case in particular is not supported by `yup-oauth2`, and even if it was, it would not be reasonable to support using the `Authenticator` struct or `GetToken` trait.

This is mainly because the server needs to store state about the user session, which requires redirecting the user to predefined callback URLs. `yup-oauth2` handles this by spinning up a server and blocking until the user response is received - entirely inappropriate for production use.

This makes it difficult to use the same credentials for multiple APIs (don't re-auth the user to use Google Photos / Google Drive - doing this manually and passing the shared token to both hubs would work).

Finally, in a server context, you need to define the oauth2 callback URL and the necessary routes to your webserver, and then manually implement the redirects and oauth2 session state (or use a middleware library). This is especially not supported by `yup-oauth2`.

google-apis-common/src/lib.rs
philippeitis: This is a bit awkward, since I think that the return type should either be `Result<String, Err>` or `Option<String>`. I think `Result<String, Err>` makes sense to make debugging easier.

There are only really two use-cases for `Result<Option<_>, _>`:
1. You want to use your own `Delegate::token` implementation without describing an error, so you return `Ok(None)` (not really important)
2. You want to use `NoToken`. However, calling `get_token()` on `NoToken` likely indicates an incorrect implementation - either the library is calling `get_token` in a no-auth context, or the user is using `NoToken` in an auth context

On further thought, I feel like `NoToken::get_token()` should either panic (not entirely unreasonable, since calling it is an implementation bug) or return an error.

philippeitis: After thinking about `Result<Option<_>, _>` a bit further, the main use-case for something like this is in APIs which may or may not require authentication based on the resource being accessed. 

For example, you might want to fetch metadata associated with publicly available Google Drive files. In this case, `Ok(None)` should indicate that the API request should proceed without a token (and any errors that result should be bubbled up). Supporting this would require some adjustments to how tokens are used, but should easy to do.

The other case that would be useful (barely) is if a user wants to call `Delegate::token` without defining their own error type. However, this doesn't seem like a particularly important case, and interferes with the other case described above, which is entirely possible.

Otherwise, if there / are no APIs which may or may not require an API key, then it would make sense for `NoToken::get_token` to simply return an error (or panic, as calling it would be an implementation bug).

philippeitis: Example usecase for `Result<Option<_>, _>`: https://issuetracker.google.com/issues/168687448?pli=1

philippeitis: Concrete usecase for `Result<Option<_>, _>` - blogger v3: https://developers.google.com/blogger/docs/3.0/reference/blogs/get

A request does not require authentication for public resources

google-apis-common/src/auth.rs
```diff
@@ -0,0 +1,163 @@
...
+//!     }
+//! }
+//! ```
+//! [`oauth2`]: https://docs.rs/oauth2/latest/oauth2/
+//! [`AccessToken`]: https://docs.rs/oauth2/latest/oauth2/struct.AccessToken.html
+//! [`Authenticator`]: yup_oauth2::authenticator::Authenticator
+use std::future::Future;
+use std::pin::Pin;
+
+type TokenResult = Option<String>;
```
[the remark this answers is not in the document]
philippeitis: That's fair enough, although with the suggested changes, the function signatures do become somewhat unwieldy.

I do want to use a type alias (maybe `GetTokenOutput` for `Pin<Box<...>>`) to keep the function signatures easier to read internally (and slightly easier to read externally?)

Byron (approved):
Thanks for detailed comments and for the adjustments, now the PR looks great to me and it's ready to merge.

Also I am looking forward to seeing what else you have planned, call me excited :).

----

Note that I only ran `make cargo-api ARGS=check` for a couple of minutes and it seemed to all look good.

PR #384, opened 2022-10-19
Update mime

philippeitis:
I noticed that the `mime` dependency is very old, and every crate the apis use which use `mime` use a more recent version.

However, this is a breaking change, going from `mime v0.2.6.0` to `mime v 0.3.16.0`.

### DIFF
--- a/google-apis-common/src/lib.rs
+++ b/google-apis-common/src/lib.rs
@@ -17,7 +17,7 @@ use hyper::header::{HeaderMap, AUTHORIZATION, CONTENT_LENGTH, CONTENT_TYPE, USER
 use hyper::Method;
 use hyper::StatusCode;
 
-use mime::{Attr, Mime, SubLevel, TopLevel, Value};
+use mime::Mime;
 
 use serde_json as json;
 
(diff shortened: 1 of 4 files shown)


google-apis-common/src/lib.rs
```diff
@@ -333,6 +333,13 @@ pub struct MultiPartReader<'a> {
 }
 
 impl<'a> MultiPartReader<'a> {
+    // TODO: This should be an associated constant
+    /// Returns the mime-type representing our multi-part message.
+    /// Use it with the ContentType header.
+    pub fn mime_type() -> Mime {
+        Mime::from_str(&format!("multipart/related;boundary={}", BOUNDARY)).expect("valid mimetype")
```
Byron, line 340: It's a bit strange that this can't be done statically. Does an API exist to build any mime type, or is string parsing for convenience?
Ultimately I don't mind as I don't think it has any impact on performance, maybe it's more readable even than the related API usage (if it exists).

Byron (approved):
Thanks a lot! Now is the time for breaking changes, and it does look good to me.

Please feel free to merge at your convenience (with a merge commit), and let me know if a new release of drive3 is desirable so I can try out the new naming scheme with a `beta-1` suffix. Thank you.

google-apis-common/src/lib.rs
philippeitis: I also thought the same, but `mime` isn't being actively maintained, and lacks quite a few features, including an easy way to do compile time `mime`.

PR #387, opened 2022-10-20
Refactor params into google-apis-common

philippeitis:
This refactors the shared url parameter functionality into `google-apis-common`. 

This is good for testing, and reduces the amount of mako-specific code-gen, which is nice. This also reduces the size of the generated `compute1` library by 2.4%, from 168MB to 164MB (and compile times from 43s to 35s, though I don't know if this is due to other effects).

Additionally, I noticed that the `url` crate is very out of date - however, I have no idea where to update this dependency, as my initial grep failed to find anything.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -7,6 +7,14 @@ on:
     branches: [main]
 
 jobs:
+  clippy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v2
+      - name: Run clippy
+        run: |
+          cargo clippy -- -D warnings
   build-and-test:
     runs-on: ubuntu-latest
     env:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,9 +25,11 @@ jobs:
       - name: Run tests
         run: |
           source ~/.profile
+          cargo test
           make test-gen
           make gen-all-cli cargo-api ARGS=test
           make cargo-api ARGS='check --no-default-features'
+          make cargo-api ARGS=check
           make cargo-api ARGS=doc
+          make cargo-cli ARGS=check
           make docs-all
-          cargo test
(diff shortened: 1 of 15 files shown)


Byron (commented):
> Additionally, I noticed that the url crate is very out of date - however, I have no idea where to update this dependency, as my initial grep failed to find anything.

You are probably looking for [this driver](https://github.com/Byron/google-apis-rs/blob/b20e63072317775e239b2400099ce1705f1c076c/etc/api/type-api.yaml#L28:L34), which is used in the shared `Cargo.toml.mako` file to add API specific dependencies. The same exists for the CLI.

Byron (approved):
I love the direction this is heading in and value this contribution greatly.

Except for some comments I think this is good for a merge, with the added note that I am usually rely on CI and don't fully understand every detail of what this is doing anymore. That's alright to me.

But, in order to help prevent regressions a little more I added `clippy` to CI as well as builds of the CLI code to prevent regressions in future. It turns out that the previous PRs introduced regressions causing the CLI to fail building, and I think it's valuable to prevent that from happening. At worst, these CLIs are an elaborate test to, at best they make it easy to test-run some APIs of interest, or even use them in scripts. The latter was the original intent.

Despite having put some breakage on display by the aforementioned changes that I pushed into this PR, I don't mean these have to be fixed here as all of it was broken before. Please feel free to merge when ready, but I strongly encourage to fix CI before continuing with any other work. Thanks for your consideration.

philippeitis:
I assumed that CI covered both the CLI and API crates in full, and was using it as such to verify changes more thoroughly (I usually run on a smaller subset of crates locally). However, it seems like it does that now, so that's good - I also appreciate the extra clippy lints.

Byron:
It's so awesome to see this fixed already, thanks so much!

>  assumed that CI covered both the CLI and API crates in full, and was using it as such to verify changes more thoroughly (I usually run on a smaller subset of crates locally). 

I fell for that too and was quite surprised when I realized it's not the case or not anymore. That said, it's worth noting that CI only runs on `youtube3` right now, and I [invite you to add more](https://github.com/Byron/google-apis-rs/blob/main/etc/api/api-list_travis.yaml#L1:L1) as you see fit.

> I also think it would be a good opportunity to push a beta release for the drive3 crate, and I'll take the time to do some more thorough testing on my own application.

Appreciated, I will make a new release of `drive3` after merging this one.

Byron:
A new release of drive3 is now [available](https://crates.io/crates/google-drive3/5.0.2-beta-1+20220225).

PR #395, opened 2022-11-29
Update yup-oauth2 to 8.0.0

fiadliel:
yup-oauth2 supports impersonating service accounts in 8.0, which is a very useful feature.
Here's an attempt to bump the dep in google-apis-common.


Byron:
Thanks a lot, much appreciated!

PR #396, opened 2022-11-29
Improve handling of plurals in generator.

fiadliel:
The generator tries to turn plural into singular nouns. However, some words turn out badly, including:
 - Addresse (from Addresses)
 - Prefixe (from Prefixes)
 - Bookshelve (from Bookshelves)
 - Sery (from Series)

By using a library, we can get a slightly better outcome. I've added an exception for "Data", because seeing that turned into "Datum" is a bit jarring.

For the above, the resulting generated word:
- Address
- Prefix
- Bookshelf
- Series

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,4 @@ codecov
 ghp-import
 pyright
 types-PyYAML
+inflect
--- a/src/generator/lib/util.py
+++ b/src/generator/lib/util.py
@@ -2,6 +2,7 @@ import os
 import re
 import subprocess
 
+import inflect
 from dataclasses import dataclass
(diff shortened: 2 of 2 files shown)


Byron:
Thanks a lot, a small but very impactful changes particularly for the books1 API !

This looks like a breaking change but I think can be part of the next releases as most of the crates have not yet been republished under the new version. In other words, another major version upgrade won't be needed.

fiadliel:
Thanks for looking! Yeah I came across `Addresse`, and it annoyed me 😄  In general, it feels most understandable here if either everything (except data) is converted to the correct singular noun, or nothing is.

PR #397, opened 2022-12-03
Remove old time dependency from API.

fiadliel:
Chrono currently depends on an old version of time with a reported vulnerability: https://rustsec.org/advisories/RUSTSEC-2020-0159

While it does not use any vulnerable code, the dependency may show up in code vulnerability scans, etc.

This removes the "oldtime" feature from chrono, to remove that. Also removes the "std" feature because it doesn't seem to be in use in this code.

### DIFF
--- a/google-apis-common/Cargo.toml
+++ b/google-apis-common/Cargo.toml
@@ -22,7 +22,7 @@ serde_with = "2.0.1"
 serde_json = "^ 1.0"
 
 base64 = "0.13.0"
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot! 

I hope [the release as v5.02](https://github.com/Byron/google-apis-rs/releases/tag/google-apis-common-v5.0.2) which includes a bump of yup-oauth2 to v8 won't break anything.

Issue #336, opened 2022-05-12
Can't set apns parameters with google-fcm1


pognibene:
Hi,
In the generated code, ApnsConfig::payload is an Option<HashMap<String, String>>.
However, according to Apple/fcm specs, this is the payload we should send:
```
"apns": {
    "fcmOptions": null,
    "headers": {
      "apns-priority": "10"
    },
    "payload": {
      "aps": {
           "sound":"sound.wav",
           "alert": {
               "loc-key": "my_loc_key"
           }
      }
    }
  }
```

I can't find a way to generate the proper apns payload with a String value.
Is this a generation bug, or I overlooked a way to achieve the desired result?
Thank you for your help!

tyranron:
@Byron I guess this may be closed now, as should be fixed by #404.

Byron:
Thanks for the hint! Done.

PR #404, opened 2022-12-30
Map 'any' JSON Schema type to json::Value instead of String

### LINKED ISSUE #403: 'any' JSON Schema type should be mapped to a json::Value instead of a String
Related to #336

Currently there is a `TODO`  stating:
```
# TODO: Figure out how to handle it. It's 'interface' in Go ...
````
(https://github.com/Byron/google-apis-rs/blob/main/src/generator/lib/types.py#L33)

I think the only suitable type here is `json::Value`.    
Using `String`  makes it it impossible to build some payloads, as mentioned in #336.

I would send a PR myself, but, for some reason, after regenerating the code I am getting a diff with size of a half of the project :smiling_face_with_tear:  And the changes are mostly unrelated.

Anyway, changing `String` to `json::Value` and regenerating the code should be sufficient.

xDarksome:
Resolves #403 #336

This PR:
- changes `any` JSON Schema type mapping to `json::Value`
- regenerates `fcm1` crate

### DIFF
--- a/gen/fcm1/Cargo.toml
+++ b/gen/fcm1/Cargo.toml
@@ -4,12 +4,12 @@
 [package]
 
 name = "google-fcm1"
-version = "4.0.1+20220228"
+version = "5.0.2-beta-1+20220228"
 authors = ["Sebastian Thiel <byronimo@gmail.com>"]
 description = "A complete library to interact with Firebase Cloud Messaging (protocol v1)"
 repository = "https://github.com/Byron/google-apis-rs/tree/main/gen/fcm1"
 homepage = "https://firebase.google.com/docs/cloud-messaging"
-documentation = "https://docs.rs/google-fcm1/4.0.1+20220228"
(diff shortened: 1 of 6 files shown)


Byron:
Thanks a lot! It looks like a change that can be very useful, and thus far it seems to work. Please note that I removed the regenerated code from the commit - PRs are much easier to review and handle that way.

Once I have run it against all existing APIs without failure, this will be merged. Would you like a new release of the `fcm1` crate once that happens?

Byron:
Closing as the commit was merged, after squashing it so GitHub can't detect this.

xDarksome:
@Byron 

> Once I have run it against all existing APIs without failure, this will be merged. Would you like a new release of the fcm1 crate once that happens?

Yes, that would be great, thanks!

Byron:
Alright, [here is new release](https://crates.io/crates/google-fcm1/5.0.2-beta-1+20220228).

PR #411, opened 2023-01-25
update apis

happysalada:
I was curious about the process of updating and so I've tried it.

One weird thing is that the speech_beta appears to not have any updates, even though the previous one is failing. I might have missed a step.

here is what I did
- `rm -f .api.deps .cli.deps && FETCH_APIS=1 make update-json -j8`
- `make cargo-api ARGS=check`
and verify that everything compiled.

As a sidenote, feel free to reject this PR if anything doesn't fit any standards. I thought this might help, but if it makes your process slower, feel free to ignore.


Byron:
Thanks for trying, it's appreciated. Unfortunately we raced for this PR should already be available in `main` by now. 

In future, please do only submit the changes to `etc/api/shared.yml`, the rest I will reproduce locally. The reason for this is that I can't review or vouch for the correctness of hundreds of JSON files, so I have to pull them myself from the authority which is the google discovery API.

PR #427, opened 2023-04-19
Match version of hyper-rustls used by yup-oauth2 8.2.0

mihirsamdarshi:
I ran cargo update in my crate and I noticed I was getting compile errors due to the versions of `hyper-rustls` used by the generated API crates don't match the version of `hyper-rustls` used by the `yup-oauth2` crate version `8.2`

### DIFF
--- a/google-apis-common/Cargo.toml
+++ b/google-apis-common/Cargo.toml
@@ -25,7 +25,7 @@ base64 = "0.13.0"
 chrono = { version = "0.4.22", default-features = false, features = ["clock", "serde"] }
 url = "= 1.7"
 
-yup-oauth2 = { version = "^ 8.0", optional = true }
+yup-oauth2 = { version = "^ 8.2", optional = true }
 itertools = "^ 0.10"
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot for bringing this up!

It seems that `yup-oauth2` incremented its minor version which contained (breaking) upgrades to `hyper-rustls`. Thus, `cargo-update` picked it up even though it ends up breaking in a program that still uses an older version of `hyper-rustls`.

So it seems that patch release of the common crate will fix this for everyone who has the latest `yup-oauth2`, but will break everybody else who doesn't use the latest. On the bright side, this case shouldn't really happen as `cargo update` would update everything at once, so… should be safe 🙏.

Byron:
Actually, I changed my mind and instead released `google-apis-common` that pins the `yup-oauth2` version to 8.1, which comes with the correct version of `hyper-rustls`. As `cargo update` won't upgrade breaking releases, `cargo update` should do the right thing in future.

PR #431, opened 2023-06-15
Switch to async io

MOZGIII:
Mako doesn't work, instructions unclear, send help. Also, 5 am here. Will resume this tomorrow..?

The basic approach is established, the rest shouldn't be too difficult.

### DIFF
--- a/google-apis-common/src/lib.rs
+++ b/google-apis-common/src/lib.rs
@@ -6,7 +6,7 @@ pub mod url;
 use std::error;
 use std::error::Error as StdError;
 use std::fmt::{self, Display};
-use std::io::{self, Cursor, Read, Seek, SeekFrom, Write};
+use std::io::{self, Cursor, Read, SeekFrom, Write};
 use std::str::FromStr;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


MOZGIII:
Turns out the vendored `virtualenv` has a bug that breaks venv on my system. I recommend updating.

Byron:
Thanks for getting started in this, it would indeed be great if these blocking calls could go away!

> Turns out the vendored `virtualenv` has a bug that breaks venv on my system. I recommend updating.

Indeed, the `makefile` installs `virtualenv` in a particular version for the user, which I consider bad practice as well. Maybe you can fix this in another PR? I don't know how portable `virtualenv` is but I'd expect the worst, but it's probably better than unconditionally installing python packages on the callers system.

MOZGIII:
I'm not sure I want to tackle this at this time. However, I'd suggest using nix (https://nixos.org/) for keeping the build environment up to date. It is portable, and can really be an opt-in, instead of being a mandatory part of the `Makefile`.
The benefit of nix is that it can really elegantly manage all the dependencies, including the python itself. The downside it it won't work on Windows.

Byron:
I am sorry, but think it's best to close this PR due to inactivity. Please feel free to reopen or resubmit.

Thanks for your understanding.

PR #434, opened 2023-06-28
Make the MessageBodyPart struct deserializable from local cache

Yatekii:


google-apis-common/src/serde.rs
```diff
@@ -160,8 +162,8 @@ pub mod urlsafe_base64 {
         where
             D: Deserializer<'de>,
         {
-            let s: &str = Deserialize::deserialize(deserializer)?;
-            base64::decode_config(s, base64::URL_SAFE).map_err(serde::de::Error::custom)
+            let s: Cow<'de, String> = Deserialize::deserialize(deserializer)?;
```
Byron, line 165: Can you elaborate why this is necessary?

gen/gmail1/Cargo.toml
```diff
@@ -23,15 +23,14 @@ mime = "^ 0.3.0"
 serde = { version = "^ 1.0", features = ["derive"] }
 serde_json = "^ 1.0"
 itertools = "^ 0.10"
-google-apis-common = { path = "../../google-apis-common", version = "5.0.1" }
-hyper = "^ 0.14"
+google-apis-common = { path = "../../google-apis-common", version = "6" }
```
Byron, line 26: Please remove all changes to this file as it is generated (and says `# DO NOT EDIT!` in its first line).
Also, please refrain from adding features to `hyper` as that is something the application will control, it's not up to the library to select features unless they are needed for it to function, which I believe isn't the case here.

As you see [in the manifest source file](https://github.com/Byron/google-apis-rs/blob/209d70bfede6a4d0a3d75064fbd2f17d395d0095/src/generator/templates/Cargo.toml.mako#L37), the library is already upgraded, thus the code needs to be regenerated and probably republished to make it come through.

Byron (changes requested):
Thanks for contributing.

In order to merge, I will need to understand more though.

gen/gmail1/Cargo.toml
Yatekii: Aw shoot, yes, I will remove it. This was the lazy hack to make it work locally :)

google-apis-common/src/serde.rs
Yatekii: So when you try to deserialize from a file reader, serde is not happy and says it expected an owned string.

Yatekii:
Sorry for the low effort PR ... will fix :)

Byron:
I am sorry, but think it's best to close this PR due to inactivity. Please feel free to reopen or resubmit.

Thanks for your understanding.

Issue #445, opened 2023-10-05
Secret manager bad decode base64 of data


omer-shtivi:
When using `secrets_versions_access` api, the serde tries to deserialize it as urlsafe, for some secrets it doesn't work as expected.
error:
`Invalid byte 43, offset 19. at line 4 column`
Unfortunately, I can't provide the full output as it contains the secret.
Here is the offending line:
https://github.com/Byron/google-apis-rs/blob/eee899d1f8d19757d7f8d50738b2070a3ecea361/gen/secretmanager1/src/api.rs#L732-L733

Confirmed from version: 5.0.2+20230114

Byron:
There are a lot of reports with problems due to incorrect deserialization of fields, which was recently adjusted to make one other API work.
It should be possible to adjust this per API.

In the mean time, it might be possible to use older versions of the crate that don't have the change, or to vendor the code and apply a patch yourself to adjust the serialization scheme.

Issue #444, opened 2023-10-02
Failed to deserialize messages from pubsub


omer-shtivi:
Crate version: 5.0.3
serde_json = "1.0.107"
serde: 1.0.188
Seems like PullResponse is unable to deserialize the message

Getting the following error:
fail to fetch message from pubsub PubsubError("JsonDecodeError(\"{\\n  \\\"receivedMessages\\\": [\\n    {\\n      \\\"ackId\\\": \\\"RVNEUAYWLF1GSFE3GQhoUQ5PXiM_NSAoRRELUxNRXHUeSBBgXzN1B1ENGXN1MCA5UkFSBxBUdF9RGx9ZXESx_4iVL1BdaXZoUxAJCkNbfF1dGgprX3SJlJWirMqMV29WYtf5291lXpbVvOZZZiE9XBJLLD5-KzJFQV5AEkwgB0RJUytDCypYEU4EISE-MD4\\\",\\n      \\\"message\\\": {\\n        \\\"data\\\": \\\"data\\\",\\n        \\\"attributes\\\": {\\n          \\\"destination\\\": \\\"dev-gcp-us-east1-omer1\\\",\\n          \\\"publisher_pod\\\": \\\"satori-management-server-api-69764c4d4c-gqhb9\\\",\\n          \\\"publisher_project\\\": \\\"stage-1-258821\\\"\\n        },\\n        \\\"messageId\\\": \\\"8329388682261733\\\",\\n        \\\"publishTime\\\": \\\"2023-10-01T18:41:37.770Z\\\"\\n      }\\n    }\\n  ]\\n}\\n\", Error(\"Invalid byte 43, offset 85.\", line: 6, column: 614))")

Byron:
This is most certainly related to #442, which in turn is related to a change that adjusted the decoding of certain fields so it works for one API, but apparently breaks it for many others. This can, and should, be enabled or handled on a per-API basis, which is possible but needs contribution.

Issue #442, opened 2023-09-18
Google TTS no longer works.


xd009642:
It doesn't matter what the request is we get responses like this:

```
Invalid byte 47, offset 18110. at line 3 column 1: {
[...]

PR #446, opened 2023-10-06
Attempt to add 'standard' base64 bytes support

emarcotte:
https://github.com/Byron/google-apis-rs/issues/442, https://github.com/Byron/google-apis-rs/issues/445, and https://github.com/Byron/google-apis-rs/issues/444 all flag an issue where some APIs respond with non-valid base64 bytes values for the "URL safe" flavor of configuration.

This adds support for a "standard" wrapper adjacent to the URL safe one with the intention of finding a way to flag which structures should use which configuration.

I'm still trying to figure out how to properly test this and I have no clue how we could mark some as URL safe and some as standard. Open to any tips!

### DIFF
--- a/google-apis-common/src/serde.rs
+++ b/google-apis-common/src/serde.rs
@@ -136,6 +136,36 @@ pub mod duration {
     }
 }
 
+pub mod standard_base64 {
+    use serde::{Deserialize, Deserializer, Serializer};
+    use serde_with::{DeserializeAs, SerializeAs};
+
+    pub struct Wrapper;
+
+    pub fn to_string(bytes: &Vec<u8>) -> String {
+        base64::encode_config(bytes, base64::STANDARD)
+    }
+
+    impl SerializeAs<Vec<u8>> for Wrapper {
+        fn serialize_as<S>(value: &Vec<u8>, s: S) -> Result<S::Ok, S::Error>
+        where
+            S: Serializer,
+        {
+            s.serialize_str(&to_string(value))
+        }
+    }
+
+    impl<'de> DeserializeAs<'de, Vec<u8>> for Wrapper {
+        fn deserialize_as<D>(deserializer: D) -> Result<Vec<u8>, D::Error>
+        where
+            D: Deserializer<'de>,
+        {
+            let s: &str = Deserialize::deserialize(deserializer)?;
+            base64::decode_config(s, base64::STANDARD).map_err(serde::de::Error::custom)
+        }
+    }
+}
+
 pub mod urlsafe_base64 {
     use serde::{Deserialize, Deserializer, Serializer};
     use serde_with::{DeserializeAs, SerializeAs};
@@ -172,7 +202,7 @@ pub fn datetime_to_string(datetime: &chrono::DateTime<chrono::offset::Utc>) -> S
 
 #[cfg(test)]
 mod test {
-    use super::{duration, urlsafe_base64};
+    use super::{duration, urlsafe_base64, standard_base64};
     use serde::{Deserialize, Serialize};
     use serde_with::{serde_as, DisplayFromStr};
 
@@ -185,8 +215,15 @@ mod test {
 
     #[serde_as]
     #[derive(Serialize, Deserialize, Debug, PartialEq)]
-    struct Base64Wrapper {
+    struct Base64URLSafeWrapper {
         #[serde_as(as = "Option<urlsafe_base64::Wrapper>")]
+        bytes: Option<Vec<u8>>,
+    }
+
+    #[serde_as]
+    #[derive(Serialize, Deserialize, Debug, PartialEq)]
+    struct Base64StandardWrapper {
+        #[serde_as(as = "Option<standard_base64::Wrapper>")]
         bytes: Option<Vec<u8>>,
     }
 
@@ -258,24 +295,45 @@ mod test {
     }
 
     #[test]
+    fn standard_base64_de_success_cases() {
+        let wrapper: Base64StandardWrapper =
+            serde_json::from_str(r#"{"bytes": "cVhabzk6U21uOkN+MylFWFRJMVFLdEh2MShmVHp9"}"#).unwrap();
+        assert_eq!(Some(b"qXZo9:Smn:C~3)EXTI1QKtHv1(fTz}".as_slice()), wrapper.bytes.as_deref());
+    }
+
+    #[test]
     fn urlsafe_base64_de_success_cases() {
-        let wrapper: Base64Wrapper =
+        let wrapper: Base64URLSafeWrapper =
             serde_json::from_str(r#"{"bytes": "aGVsbG8gd29ybGQ="}"#).unwrap();
         assert_eq!(Some(b"hello world".as_slice()), wrapper.bytes.as_deref());
     }
 
     #[test]
     fn urlsafe_base64_de_failure_cases() {
-        assert!(serde_json::from_str::<Base64Wrapper>(r#"{"bytes": "aGVsbG8gd29ybG+Q"}"#).is_err());
+        assert!(serde_json::from_str::<Base64URLSafeWrapper>(r#"{"bytes": "aGVsbG8gd29ybG+Q"}"#).is_err());
+    }
+
+    #[test]
+    fn standard_base64_de_failure_cases() {
+        assert!(serde_json::from_str::<Base64StandardWrapper>(r#"{"bytes": "%"}"#).is_err());
     }
 
     #[test]
     fn urlsafe_base64_roundtrip() {
-        let wrapper = Base64Wrapper {
+        let wrapper = Base64URLSafeWrapper {
             bytes: Some(b"Hello world!".to_vec()),
         };
         let s = serde_json::to_string(&wrapper).expect("serialization of bytes infallible");
-        assert_eq!(wrapper, serde_json::from_str::<Base64Wrapper>(&s).unwrap());
+        assert_eq!(wrapper, serde_json::from_str::<Base64URLSafeWrapper>(&s).unwrap());
+    }
+
+    #[test]
+    fn standard_base64_roundtrip() {
+        let wrapper = Base64StandardWrapper {
+            bytes: Some(b"Hello world!".to_vec()),
+        };
+        let s = serde_json::to_string(&wrapper).expect("serialization of bytes infallible");
+        assert_eq!(wrapper, serde_json::from_str::<Base64StandardWrapper>(&s).unwrap());
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


emarcotte:
I ran the regenerate process and pushed everything to https://github.com/emarcotte/google-apis-rs/tree/regenerate, then pointed a contrived program to it which was able to successfully read a secret that would not decode as url-safe.

Byron:
Fantastic! I think the solution could be *(instead of relying to have per-API settings for this which need to be created by hand in absence of a way to infer this)* to have a wrapper that *tries* both encodings and fails only once both attempts fail.

That way we only have to choose the most common encoding to be tried first in order to avoid too many wasted cycles (even though it would probably detect early when the encoding doesn't pan out so failure is cheap anyway).

What do you think?

emarcotte:
I'll give it a shot sometime this week. I'm not an expert in base64 so my only question would be if there's any possible scenario where we do an incorrect, but non-erroring parse. 

BTW; any hints on that clippy lint?

Byron:
Thanks for your help!

> I'll give it a shot sometime this week. I'm not an expert in base64 so my only question would be if there's any possible scenario where we do an incorrect, but non-erroring parse.

Me neither, but: the question here is if decoding could succeed by one but produce garbage. And I'd think the answer for this is strictly: no. Of course, this might just be wishful thinking, but I am inclined to believe it's true and not worry about it.

> BTW; any hints on that clippy lint?

Each time the Rust version increments `clippy` learns new tricks, which can lead to a failing CI once the runner gets the upgrade.
You can probably easily reproduce it locally in the crate itself (it's not generated) and fix it as suggested by clippy itself.

The none-lint failure is more of a problem though as it seems python related (Oh I have so learned my lesson when it comes to python: do not use for anything that has to last).

Thanks again.

emarcotte:
On the "fetching" side I think a fallback approach can work but I am not so confident that on the sending side. If say PubSub msg payloads need to be URL safe then we'd be posting invalid data which is a bit less than burning cycles in decoding.

Would you expect that we only do the fallback on reading side? 

BTW I did finally realize google standard APIs [_CLAIM_ to be URL safe](https://developers.google.com/discovery/v1/type-format) variant but clearly that isn't followed very precisely (or the docs are bad).

Byron:
> On the "fetching" side I think a fallback approach can work but I am not so confident that on the sending side.

Right, I agree! The fallback would only work for the reading side.
Getting it to work for sending would mean there is a mechanism for the user to configure this, so the user would have to figure it out anyway. And if that's the case, whatever the user configures might as well be used for the receiving end and avoid having to guess anything.

The question really is if the user has a way to know. Is there something in the API description that indicates what is what? I am pretty sure that the generators that Google does provide for other programming languages have figured this out, so maybe one could learn from there?

emarcotte:
Haven't forgotten this, just been swamped. Been trying to research how go and python clients work and mostly just getting more confused how this fixed anything and why the original code didn't work...

Byron:
>  just getting more confused how this fixed anything and why the original code didn't work..

I think this is related to *some* APIs acting differently, but I also never validated this myself. So me it seemed the fix broke more than it fixed and maybe this tells use we should revert first and then try to fix the original issue another way.

If you agree, please feel free to open a revert-PR for that.

andrewbaxter:
Sorry, just dropping in.  I assume the issue is random apis use random different base64 standards, and it's not part of the metadata?  It could be like 50/50 urlsafe vs standard?

What about one of the following:
1. A base64 decoder that tries both standard and urlsafe? Or tries to normalize first (swapping underscores or whatever out)?  Is this just decoding? Maybe gcp is permissive in which b64 it accepts?
2. Maintain a separate list of per-api metadata including which base64 style is used.  Making MRs for individual apis which have issues as I (we?) encounter them doesn't seem too hard.  The initial metadata could be 100% base64.

andrewbaxter:
Uhh, so AFAICT the golang go library uses the standard go `json` package, which always uses standard (not web-safe) base64 encoding and decoding.  This directly contradicts their own guidance but well

Byron:
Maybe it can just be switched back to that then. From my experience, after it changed to what it is now the number of issues multiplied - so what was a fix for few turned out to be a problem for many.

Maybe you can find the commit and revert it, which should then be standard base64 as well.

Alternatively, figure out a way to let the user handle encoding and decoding to a greater extend.

By the way, I am very sorry you all have to suffer through this and my hands-off approach to the project. Too many people use it and it's in a barely-community-maintained state - right now it's not good enough. I simply can't understand why Google doesn't provide their own bindings for Rust.

andrewbaxter:
Looking at python, there's two bytes handlers registered here: https://github.com/googleapis/proto-plus-python/blob/161f89f5b7a6f493602ed8890ee9a11ebbc77836/proto/marshal/marshal.py (I think this is the library python uses)

The first one comes from googles protobuf library, and I _think_ the "bytes" type has id 18.  The 2nd one is defined locally, with id 12.  I think maybe the python methods might be automatically generated based on protobuf specs?

The first one uses standard for encode, and urlsafe for decode :question: https://github.com/protocolbuffers/protobuf/blob/a7b0421c78412baa48704d727601a17ac0f451d1/python/google/protobuf/json_format.py AFAICT ... that's using id type 12 though.  Maybe I read the 18 thing wrong.

The 2nd one uses urlsafe.

andrewbaxter:
One wonders why google does a lot of things

andrewbaxter:
And this is way better still than writing it from scratch, especially if you don't mind reviewing and merging!

andrewbaxter:
I'm going to try some read/write api with a couple libraries and see what goes over the wire.

andrewbaxter:
So doing a file upload with checksum=crc32c

Golang sends
```
"crc32c":"ksgKMQ=="
```
receives:
```
"Etag":[]string{"CNGDo77Aw4MDEAE="},
...
"md5Hash":"4vxxTEcn7pOV8yTNLn8zHw==","crc32c":"ksgKMQ==","etag":"CMTc+/Gzw4MDEAE=",
```

Python sends
```
"crc32c": "ksgKMQ=="
```
receives
```
header: ETag: CPic/oa8w4MDEAE=
```
but no body because apparently in python it's impossible to log http response bodies.

There _is_ a lot of explicit encoding/decoding though in both, stuff that looks like it wasn't automatically generated (from protobuf or otherwise).  The checksum is like that, so it might not be a good example.

andrewbaxter:
As someone who seems deeply familiar with implementing GCP api libraries, if you don't mind a totally out of left-field ping @codyoss - would you be able to lend advice on what base64 encoding is canonical here? (or maybe confirm that it's a case-by-case basis thing)

Otherwise I think:
1. Revert (or switch to standard).  I'll try to look up that commit
2. Add a mechanism for pinpoint overrides

Would be the safest...

andrewbaxter:
Was this the commit? https://github.com/Byron/google-apis-rs/pull/379 - If I'm reading that, before it was just `String` etc, so neither urlsafe nor standard base64.

In that case I'm not sure overrides make sense.  I kind of do feel reverting is safest and easiest here, and having `String` is still a far cry better than having to set up http requests manually.  If there were a clear path to decoding that would be awesome though.

Byron:
Thanks a lot for your help, I will look at the PR in a bit!

> 2. Add a mechanism for pinpoint overrides

Such a mechanism exists in principle, as each API can have overrides to any of its properties, but the implementation would have to be adjusted to use that to drive the decoding or encoding of values.

With #379 that can probably be postponed indefinitely though.

andrewbaxter:
FWIW I made #463 .  Having base64/datetime automatically parsed would be awesome but I don't have any great ideas.  Maybe if someone else wants to tackle it it could be revert-reverted.

codyoss:
@andrewbaxter  can you provide me with a little more context what are you encoding/decoding? It depends on which part of a req/resp you are dealing with which type to use.

andrewbaxter:
Oh yeah! Thanks for taking the time to respond!

I think it's _just_ the JSON request/response bodies from the APIs so it wouldn't include query parameters or auth.  The generation comes from, for example:
https://texttospeech.googleapis.com/$discovery/rest?version=v1beta1 .

https://developers.google.com/discovery/v1/type-format says
> A padded, base64-encoded string of bytes, encoded with a URL and filename safe alphabet (sometimes referred to as "web-safe" or "base64url"). Defined by [RFC4648](http://tools.ietf.org/html/rfc4648).

So the assumption was that this should be web safe base64 everywhere, but people encountered various APIs which are returning standard base64 strings.  AFAICT the golang implementation seems to contradict that and is nearly all standard base64, but there are a few things that appear to use urlsafe, and python appears to use urlsafe if I read the code correctly...

I was wondering if you could give a hint on which one is correct, or if you had hints on the correct way to deal with base64 in json in an API implementation.

So you think standard encoding would be the safest to use here?  Do you know if there are a lot of apis that deviate from that (i.e. few enough to be fixable by hand)?

codyoss:
> AFAICT the golang implementation seems to contradict that and is nearly all standard base64, but there are a few things that appear to use urlsafe, and python appears to use urlsafe if I read the code correctly...

You are correct for our Go libraries. I can't speak for the python ones -- I assume some libraries have either fall back behaviour or just handle bytes in different ways at a language level.

A good test case for handling bytes is a body is a pubsub message receive RPC.

> I think it's just the JSON request/response bodies from the APIs so it wouldn't include query parameters

Some fields laid out in discovery are query params and not a part of a body

> So you think standard encoding would be the safest to use here? 

Yes

andrewbaxter:
Thanks @codyoss ! In that case @Byron would you mind if we do this?

- Merge this MR
- I can commit to helping adding override functionality if other base64 issues appear in say the next year (i.e. if some things do need urlsafe base64)

I just generated the urlsafe code from this branch locally and I'll do some testing.  I'm not sure if there are tests that need to be fixed, it looks like the jobs expired - are you able to rerun the tests? Or I could make a new MR (unless @emarcotte wanted to continue this).

emarcotte:
Hey guys, sorry i've been absent for months, I've been completely slammed by Real Life (TM). I had intended to just go back to string value but never found time :(

andrewbaxter:
No worries! Cool, I'll take it over then if that's alright.

andrewbaxter:
I made #464 with a new build, no new changes

Byron:
Great! After all, the PR could be merged and apparently we all lost our track here in this PR somehow. Thanks @andrewbaxter for guiding us to completion.

Before re-releasing all crates I think it's worth to take another look and make small last-minute changes in the next couple of days. Then when deemed ready, I can cut the release.

Thanks everyone.

andrewbaxter:
Awesome, thank you too for help with the merge!  Sounds good to me.

PR #450, opened 2023-10-21
Add Debug derive to youtube3::api::Scope enum

OMGeeky:
Adding that makes logging and debugging easier.

### DIFF
--- a/gen/youtube3/src/api.rs
+++ b/gen/youtube3/src/api.rs
@@ -23,7 +23,7 @@ use crate::{client, client::GetToken, client::serde_with};
 /// Identifies the an OAuth2 authorization scope.
 /// A scope is needed when requesting an
 /// [authorization token](https://developers.google.com/youtube/v3/guides/authentication).
-#[derive(PartialEq, Eq, Hash)]
+#[derive(PartialEq, Eq, Hash, Debug)]
 pub enum Scope {
     /// Manage your YouTube account
     Full,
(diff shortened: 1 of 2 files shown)


OMGeeky:
I honestly don't really know how to build all things (and if I should?) but even if I should, it generates me a ton of changes with whitespace/line endings so if those should not be committed too I can't really generate it all since I would have to manually select the actual changes on each file etc.

If you got anything I should do differently feel free to tell me.

I also have no Idea why the check failed, maybe you can give me a hint for that one since it worked locally.

Byron (changes requested):
CI is once again broken because python casually breaks over time with nothing else changing, or the dependencies that are pulled in do. This is a problem on its own here.

The file edited here is generated, and changes need to be made in the generator or they will be lost next time the file is generated.

The source for it [is here](name="scope_enum()) if you want to give it another try. Files can be regenerated with `make youtube3` (hopefully), and if it doesn't regen after changing the code throw an `rm -Rf gen/youtube3` in for good measure.

OMGeeky:
I did (or at least think I did) change the generator file that is at `src/generator/templates/api/lib/lib.mako`. Btw for me at least the link you sent does not work.
This PR should include 2 files, one is the generated one with `make youtube3` and the other one is the template for it. Should I remove the generated one from the PR or is that fine?

Maybe while were at it we could also add the Clone derive to it since I don't see a reason why it should not be cloneable and the Enum does not contain any data that might not be cloneable

Byron:
Apologies, I only saw the first line which was a generated file and stopped there. 

> Should I remove the generated one from the PR or is that fine?

Yes, please do remove the changes in the generated file do feel free to add more traits now that we are at it.

The link also doesn't work for me, strangely enough, but it has no value as it just linked to the generator which you are already aware of.

Thanks again.

Byron:
Thanks a lot!

I added even more traits before merging.

PR #458, opened 2023-11-29
Update README.md

anil-e:
replaced `messages_import` with `messages_list` because [UserMessageImportCall](https://docs.rs/google-gmail1/latest/google_gmail1/api/struct.UserMessageImportCall.html#) does not provide a `doit()` method


Byron:
These are generated files, and the help is generated too. A fix would have to be made in the generator itself.

PR #465, opened 2024-01-06
Make field mask inner public

andrewbaxter:
The inner value was private so it couldn't be constructed directly (without deserializing from a string indirectly).  Creating it directly seems safer than deserializing, especially since I'm unfamiliar with the format.

Was it intended for this to be constructed directly or is there another mechanism for setting these up? I thought the api might do it directly but it doesn't atm afaict.

### DIFF
--- a/google-apis-common/src/field_mask.rs
+++ b/google-apis-common/src/field_mask.rs
@@ -32,7 +32,13 @@ fn snakecase(source: &str) -> String {
 
 /// A `FieldMask` as defined in `https://github.com/protocolbuffers/protobuf/blob/ec1a70913e5793a7d0a7b5fbf7e0e4f75409dd41/src/google/protobuf/field_mask.proto#L180`
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
-pub struct FieldMask(pub Vec<String>);
+pub struct FieldMask(Vec<String>);
+
+impl FieldMask {
+    pub fn new<S: AsRef<str>>(values: &[S]) -> Self {
+        return Self(values.iter().map(|s| snakecase(s.as_ref())).collect());
+    }
+}
 
 impl Serialize for FieldMask {
     fn serialize<S>(&self, s: S) -> Result<S::Ok, S::Error>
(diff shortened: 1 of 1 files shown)


Byron (changes requested):
It's probably easiest to allow users to just let users create such masks directly, after all they serialize into comma-separated lists like `"a,b"`

However, it turns out that these *must* be snake-case, and anything the user passes in should adhere to that.

Thus I recommend adding a constructor instead that converts to snake-case, just like the deserialization (`FromStr`) does.

andrewbaxter:
Okay sounds good.  I made the change, and tested it locally although I was already using a non-snake-case string `"state"`.

That said, as you mentioned the specification isn't exactly simple.  I think in order to use the field correctly in the first place users will have to read up on its use, and if despite doing that they're passing the wrong values in it might be best to allow the api call to fail to let them know they need to re-read.  Magically lowercasing but not checking field existence, etc might be more complex implicit behavior.

google-apis-common/src/field_mask.rs
```diff
@@ -34,6 +34,12 @@ fn snakecase(source: &str) -> String {
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub struct FieldMask(Vec<String>);
 
+impl FieldMask {
+    pub fn new<S: AsRef<str>>(values: &[S]) -> Self {
```
Byron, line 38: The public API should be documented as it [goes up on docs.rs](https://docs.rs/google-apis-common/latest/google_apis_common/field_mask/struct.FieldMask.html).

andrewbaxter:
Or were you thinking it should just be `FieldMask(pub String)`?  I wouldn't mind that, it'd convey to users that they're completely on their own which I think might be good.

Byron (changes requested):
Thanks!

The type already exhibits this behaviour and it internally wants to have snake-case - the constructor enforces this invariant and that seems fine to me.
Nothing changes otherwise, the user would always have to know valid strings for field names and there is no type-safety for that (or it isn't implemented).

google-apis-common/src/field_mask.rs
andrewbaxter: Added!

Byron (approved):
Thanks a lot for your help!

PR #471, opened 2024-02-03
Fix base64 deserialization when using a reader

ytanay:
In our use case, we deserialize GCP structs using a file reader (instead from an in-memory string), which yields errors such as this is:

```
Caused by:
    invalid type: string "...", expected a borrowed string at line ... column ...
```

The first commit demonstrates this behavior, and the second fixes it by allow `serde` to choose between a borrowed an owned string using a `Cow`.


Byron (approved):
Thanks a lot for the fix!

I didn't realise that `serde` can issue runtime errors like that, even though it makes perfect sense that it needs storage in order to deserialize a stream which wasn't taken into consideration. Quite a footgun, I will try to remember that.

PR #482, opened 2024-03-21
Update yup-oauth2 to 8.3.3 and hyper-rustls to 0.25

### LINKED ISSUE #481: Support yup-oauth2 v8.3.3 with hyper-rustls v0.25
The following has been merged and released:

https://github.com/dermesser/yup-oauth2/pull/213

IvanUkhov:
Closes #481.

### DIFF
--- a/google-clis-common/Cargo.toml
+++ b/google-clis-common/Cargo.toml
@@ -17,7 +17,7 @@ doctest = false
 
 [dependencies]
 mime = "^ 0.3"
-yup-oauth2 = "^ 8.2"
+yup-oauth2 = "8.3.3"
 serde = "1"
 serde_json = "1"
 strsim = "0.10.0"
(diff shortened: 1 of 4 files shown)


IvanUkhov:
@Byron, here is an illustration.

IvanUkhov:
OK, it works with unwrapping. But I am not sure if this is appropriate at those places.

Byron (approved):
Thanks a lot! It's great this just worked.

The unwraps are acceptable - one is visible in documentation that uses unwraps anyway, the other in the CLI which, I think, aren't used very much.
And even if they are and this fails, the unwrap will make clear enough what happened. If it's a problem, a contribution can fix that issue as well.

Definitely good to go. Please note that I won't create a new release until later this year, so using a vendered version fo the generated code might be the best way to go.

PR #485, opened 2024-04-21
Add support for Spanner

### LINKED ISSUE #484: Add support for Spanner
Hello again,

As mentioned, I am planning to work on Spanner. It is currently in the blacklist, and removing it from there shows that the first problem is duplicate type definitions. I have looked closer, and it is due to concatenation of path segments and singularization. There are distinct entries in the API that after processing this way become indistinguishable, causing the failure.

Here is the naming code involved:

https://github.com/Byron/google-apis-rs/blob/main/src/generator/lib/util.py#L1062

Before I do anything, I would like to ask you what would be the best strategy to proceed. Intuitively, I suspect that many APIs are blacklisted due to this conversion. Ideally, it should be such that it does not lead to ambiguity when the dots are removed. Here are some ideas:

* Do not singularize and see what happens. At least, in the case I looked at, it would help.
* Come up with a name that will take the place of the dot, like `Dot` (saw it in some other crate). It will be many `Dot`s in types, but it will be unambiguous.

Both will break the current interface. One would have to do bump the major version. Do you have other ideas?

Thank you!

IvanUkhov:
Hello,

Here are the changes needed for Spanner to be generated and successfully tested.

I am not sure what the protocol is. Should I include the generated package into the same PR? But then it looks like the generated bindings are outdated with respect to the JSON files in the repository (probably also related to #482), and the JSON files in the repository are outdated with respect to the official reference. It would be nice to bring everything up to seed. I can do that separately if you want.

Closes #484.

### DIFF
--- a/etc/api/api-list_travis.yaml
+++ b/etc/api/api-list_travis.yaml
@@ -1,4 +1,6 @@
 api:
   list:
+    spanner:
+    - v1
     youtube:
     - v3
--- a/etc/api/shared.yaml
+++ b/etc/api/shared.yaml
@@ -27,7 +27,6 @@ api:
     # Has a struct field called 'async', which is now reserved
     - homegraph
     - gkehub
-    - spanner
     - firebaserules
     - servicemanagement
     # defines its own `Option` type
--- a/src/generator/lib/types.py
+++ b/src/generator/lib/types.py
@@ -70,7 +70,7 @@ RUST_TYPE_RND_MAP = {
     f"{CHRONO_PATH}::Duration": lambda: f"chrono::Duration::seconds({randint(0, 9999999)})",
     CHRONO_DATE: chrono_date,
     CHRONO_DATETIME: lambda: CHRONO_UTC_NOW,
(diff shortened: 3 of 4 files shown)


Byron (changes requested):
That's great work, thanks so much for contributing!

CI actually has a separate list for what's 'all' APIs, allowing it to build and check a few representative APIs. Could you add 'spanner' to that to have it tested as the first previously impossible APIs?

Regarding the JSON file update (API descriptions), please feel free to update them in a separate PR. I was under the impression that I did that not too long ago though, but that might be wrong.

And in yet another PR, I can imagine that more of the exclude-listed-but-now-working APIs could be re-enabled, but it's certainly not required, it would just be nice to leverage the new capabilities introduced here to the fullest.

From there, I should be able to create a new release with everything up-to-date, or at least release the newly available APIs.

Byron:
Thanks for the update!

> [..] By the way, it is no longer Travis; one should probably clean up and rename what is left from the time when it was in use. There is `.travis.yml` etc. I can do that.[..]

Thanks, that would be great!

All looks good now! And I hope more APIs can be activated now that the type-generator has been improved.

PR #488, opened 2024-04-22
Update the JSON schemas

IvanUkhov:
As discussed in #485, here is a separate PR for updating the JSON schemas.


Byron (changes requested):
Thanks a lot!

Could you rewrite the commit so that it contains the command(s) that was executed? I think this might help some folks to do the same locally and/or make the results more reproducible.

Thanks again.

IvanUkhov:
I have included the executed command into the commit message, If I understood you right. Also rebased.

Byron:
Thanks a lot!

PR #490, opened 2024-04-23
Update yup-oauth2 in google-apis-common

IvanUkhov:
I have missed this in #482. The current specification still allows 8.3.3, but is good to set this lower limit still. It is not going to compile with anything lower.


Byron (approved):
Thanks a lot for your help, it's much appreciated!

Byron:
Strange - this passed here, but [after merging it failed](https://github.com/Byron/google-apis-rs/actions/runs/8811368553). Could you check and submit a PR with the fix? Thanks so much.

IvanUkhov:
Oh, they yanked that version just now! They probably realized it had a breaking change. I will update to the new version then.

PR #492, opened 2024-04-24
Update the license

IvanUkhov:


LICENSE.md
```diff
@@ -1,12 +1,7 @@
-<!---
-DO NOT EDIT !
-This file was generated automatically from 'src/generator/templates/LICENSE.md.mako'
-DO NOT EDIT !
--->
 The MIT License (MIT)
 =====================
 
-Copyright © `2015-2020` `Sebastian Thiel`
+Copyright 2015–2024 Sebastian Thiel
```
IvanUkhov, line 4: Updating to the official template:

https://opensource.org/license/MIT

Byron: 9 years?? Oh my…!

LICENSE.md
```diff
@@ -1,12 +1,7 @@
-<!---
```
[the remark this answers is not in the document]
Byron: Maybe, but even if it is generated, the next time it would at least have the correct information in it.

etc/api/shared.yaml
```diff
@@ -105,7 +105,7 @@ cargo:
 urls:
   authenticator_delegate: https://docs.rs/yup-oauth2/*/yup_oauth2/trait.AuthenticatorDelegate.html
 copyright:
-  license_abbrev: "MIT"
-  years: "2015-2020"
+  license_abbrev: MIT
+  years: 2015–2024
   authors:
-    - "Sebastian Thiel"
+    - Sebastian Thiel
```
Byron, line 111: Maybe here there should be a line with blanked "Contributors"? What do you think?

Byron (commented):
Thanks a lot!

This looks good to me, and we could even make another improvement if you agree (see comments).

LICENSE.md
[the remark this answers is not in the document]
IvanUkhov: Do you mean this file in the root is also generated automatically? Maybe I missed something. I can recover that comment at the top.

etc/api/shared.yaml
IvanUkhov: Sorry. Can you please say more? Where would you like it to appear?

Byron:
Looking at it again, I think it's OK the way it is. No need for complications, and sorry for the hassle.

PR #493, opened 2024-04-26
update dependencies

serprex:
base64 0.22, serde_with 3, url 2

url v2 no longer exports percent_encoding, add percent-encoding v2 as explicit dependency

percent-encoding v2 no longer includes predefined encoding sets

No changes necessary for serde_with or itertools


Byron:
Thanks a lot!

PR #495, opened 2024-05-06
Made doit() method for ObjectInsertCall public

Rodion-Bozhenko:
Readme for google_storage1 says that .objects().insert() method could be invoked like this 
```rust
let r = hub.objects().insert(...).doit().await
```
but it is impossible since doit() is private

### DIFF
--- a/gen/storage1/src/api.rs
+++ b/gen/storage1/src/api.rs
@@ -22890,7 +22890,7 @@ where
 
 
     /// Perform the operation you have build so far.
-    async fn doit<RS>(mut self, mut reader: RS, reader_mime_type: mime::Mime, protocol: client::UploadProtocol) -> client::Result<(hyper::Response<hyper::body::Body>, Object)>
+    pub async fn doit<RS>(mut self, mut reader: RS, reader_mime_type: mime::Mime, protocol: client::UploadProtocol) -> client::Result<(hyper::Response<hyper::body::Body>, Object)>
 		where RS: client::ReadSeek {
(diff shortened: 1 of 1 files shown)


Byron:
It's recommended to apply patches in locally vendored copies of the source code as needed.
However, it makes no sense to apply patches to a file in `gen/` as all these will be overwritten the next time the generator runs.

A viable patch would fix the documentation generator to not propose using methods which are private, but instead use their public counterparts.

IvanUkhov:
Indeed, the files in `gen` are generated programmatically. They are not supposed to be changed by hand. So if you would like to propose some change, it has to be done in the templates used for the generation. They are here:

https://github.com/Byron/google-apis-rs/blob/main/src/generator/templates

And here is the documentation that has to be updated:

https://github.com/Byron/google-apis-rs/blob/main/src/generator/templates/api/lib/lib.mako#L154-L163

And here are the lines deciding whether it is public or private:

https://github.com/Byron/google-apis-rs/blob/main/src/generator/templates/api/lib/mbuild.mako#L439-L471
https://github.com/Byron/google-apis-rs/blob/main/src/generator/templates/api/lib/mbuild.mako#L528

Rodion-Bozhenko:
> It's recommended to apply patches in locally vendored copies of the source code as needed. However, it makes no sense to apply patches to a file in `gen/` as all these will be overwritten the next time the generator runs.
> 
> A viable patch would fix the documentation generator to not propose using methods which are private, but instead use their public counterparts.

Got it. Thanks for response

PR #498, opened 2024-05-16
Fix clippy complaining about "direct implementation of `ToString`"

OMGeeky:
The clippy action currently fails due to a direct implementation fo ToString instead of Display

The error message is:
```
error: direct implementation of `ToString`
   --> google-clis-common/src/lib.rs:128:1
    |
    | / impl ToString for FieldCursor {
    | |     fn to_string(&self) -> String {
    | |         self.0.join(".")
    | |     }
    | | }
    | |_^
    |
    = help: prefer implementing `Display` instead
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#to_string_trait_impl
    = note: `-D clippy::to-string-trait-impl` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::to_string_trait_impl)]`
```

### DIFF
--- a/google-clis-common/src/lib.rs
+++ b/google-clis-common/src/lib.rs
@@ -11,9 +11,9 @@ use std::io;
 use std::io::{stdout, Write};
 use std::path::Path;
 use std::str::FromStr;
-use std::string::ToString;
 
 use std::default::Default;
+use std::fmt::Display;
 
 const FIELD_SEP: char = '.';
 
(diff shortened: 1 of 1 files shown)


Byron (approved):
Much better, thank you!

google-clis-common/src/lib.rs
```diff
@@ -125,9 +125,9 @@ impl AsRef<str> for CallType {
 #[derive(Clone, Default)]
 pub struct FieldCursor(Vec<String>);
 
-impl ToString for FieldCursor {
-    fn to_string(&self) -> String {
-        self.0.join(".")
+impl Display for FieldCursor {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0.join("."))
```
Byron, line 130: This could also be `self.0.join(".").fmt(f)` I think. In case you like it more, that could be changed in the other PR of yours.

OMGeeky: That is true, I don't really care too much I'm just used to the `write` macro way but both ways are fine with me.

Issue #502, opened 2024-06-05
List compute engine virtual machines fails to parse API response if a VM has "owner" as a label key


dantreiman:
I've written a function which uses google_compute1 5.0.4 to list virtual machines in a zone:

```rust
 let compute_engine = get_compute_engine_client().await.expect("Failed to create compute engine API client.");
 match compute_engine.instances().list(PROJECT_ID, ZONES).doit().await {
    ...
}
```

This function succeeds normally.  However, if I add a label with key `"owner"` to a VM in the specified zone, the request fails with error:  `Invalid symbol 45, offset 3. at line 92 column 40`

For example, I added the label `"owner": "dtreiman"` to a VM and the request fails.  Remove the `"owner"` label, request succeeds. 

Update: I tried using `"resource_owner" : "dtreiman"` and `"requester":"dtreiman"` and the error appears.
However, it works with `"requester":"me"` or `"foo": "bar"`

Byron:
Thanks for reporting!

I think here it would be easiest to vendor the crate (or put it locally), use a `[patch."crates-io"]` in the dependent crate to point to it, and add debugging code to see what's going on. Maybe extract the received JSON verbatim to understand why it doesn't seem to parse.

From there, a local fix should be possible, and maybe it's possible too to derive a more general fix as well.

dantreiman:
Thanks!  Per your suggestion, I cloned the crate locally so I could insert debugging code. The error appears to be happening during parsing of the response. `Invalid symbol 45, offset 8. at line 25 column 39` refers to ASCII character 45 which is `-`.

Line 25 in the response below is: `"fingerprint": "1cecahVF-n8=",`.  So, maybe this error happens when the `-` character is present in the fingerprint.

```json
{
"kind": "compute#instanceList",
"id": "<REDACTED>",
"items": [
{
"kind": "compute#instance",
"id": "1879471610156305019",
"creationTimestamp": "2024-06-06T10:40:38.876-07:00",
"name": "instance-20240606-174003",
"description": "",
"tags": {
"fingerprint": "42WmSpB8rSM="
},
"machineType": "<REDACTED>",
"status": "RUNNING",
"zone": "<REDACTED>",
"canIpForward": false,
"networkInterfaces": [
{
"kind": "compute#networkInterface",
"network": "<REDACTED>",
"subnetwork": "<REDACTED>",
"networkIP": "<REDACTED>",
"name": "nic0",
"fingerprint": "1cecahVF-n8=",
[...]

PR #506, opened 2024-06-11
Tolerant base64 decoding.  Accept url-safe or standard encoding in responses.

dantreiman:
Attempts to decode `bytes` fields in response JSON using both `standard` and `url_safe` encodings.  The expected encoding's desierializer is tried first.  On error the second is attempted.  If the second fails, the original error from the first is returned.

Fixes https://github.com/Byron/google-apis-rs/issues/502

Likely fixes other bugs too but I haven't tested

### DIFF
--- a/google-apis-common/src/serde.rs
+++ b/google-apis-common/src/serde.rs
@@ -172,7 +172,7 @@ pub mod standard_base64 {
             let s: Cow<str> = Deserialize::deserialize(deserializer)?;
             match base64::prelude::BASE64_STANDARD.decode(s.as_ref()) {
                 Ok(decoded) => Ok(decoded),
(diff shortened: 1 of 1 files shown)


Byron (changes requested):
That's great, I am so looking forward to getting this released.

Could you take a look at the failing tests and adjust them? You could even add some assertions to assure both serializers work as expected (and that one failure already indicates it does as it doesn't fail anymore).

Thanks again!

dantreiman:
Sounds good.  I've added two cases using `+`, `/` and `-`, `_` respectively to ensure decoder doesn't error out.  Changed failure case to a character outside either accepted base64 character set.

Byron (approved):
That's great, thanks so much!

PR #509, opened 2024-06-11
Bump strsim to v0.11 and drop it from generated CLI crates

paolobarbolini:
Also drops it from CLI crates which import it without using it.


Byron:
Great catch, thank you!

PR #511, opened 2024-06-15
Derive utoipa::ToSchema by default for generating openapi specs by other libraries

that-ambuj:
I needed this feature because our consuming API needs to generate docs for the frontend client but could not because of Rust's orphan rule.

This is PR is incomplete, I think it would be better to put this feature in a feature flag because everyone would not need it.

Please validate that this feature is a good idea.

### DIFF
--- a/src/generator/lib/util.py
+++ b/src/generator/lib/util.py
@@ -1199,7 +1199,7 @@ def unique(
     original: str,
     desired: str,
     attempts: int = 0,
-    assigned: dict[str, str] = {},
+    assigned: Dict[str, str] = {},
 ) -> str:
     if original in assigned:
         return assigned[original]
--- a/src/generator/templates/Cargo.toml.mako
+++ b/src/generator/templates/Cargo.toml.mako
@@ -31,7 +31,7 @@ hyper-rustls = "0.25.0"
 ## Must match the one hyper uses, otherwise there are duplicate similarly named `Mime` structs
 mime = "^ 0.3.0"
 serde = { version = "^ 1.0", features = ["derive"] }
-utoipa = { version = "4.2", features = ["axum_extras", "actix_extras", "uuid", "chrono", "url"] }
+utoipa = { version = "^4.2", features = ["axum_extras", "uuid", "chrono", "url"] }
 serde_json = "^ 1.0"
 itertools = "0.13"
 % if cargo.get('is_executable', False):
--- a/src/generator/templates/Cargo.toml.mako
+++ b/src/generator/templates/Cargo.toml.mako
@@ -31,7 +31,7 @@ hyper-rustls = "0.25.0"
 ## Must match the one hyper uses, otherwise there are duplicate similarly named `Mime` structs
 mime = "^ 0.3.0"
 serde = { version = "^ 1.0", features = ["derive"] }
-utoipa = { version = "^4.2", features = ["axum_extras", "uuid", "chrono", "url"] }
+utoipa = { version = "^4.2", optional = true }
 serde_json = "^ 1.0"
 itertools = "0.13"
 % if cargo.get('is_executable', False):
@@ -62,4 +62,5 @@ version = "${util.crate_version()}"
 [features]
 yup-oauth2 = ["google-apis-common/yup-oauth2"]
 default = ["yup-oauth2"]
+utoipa = ["dep:utoipa", "utoipa/url"]
 % endif
--- a/src/generator/templates/Cargo.toml.mako
+++ b/src/generator/templates/Cargo.toml.mako
@@ -31,6 +31,7 @@ hyper-rustls = "0.25.0"
 ## Must match the one hyper uses, otherwise there are duplicate similarly named `Mime` structs
 mime = "^ 0.3.0"
 serde = { version = "^ 1.0", features = ["derive"] }
+utoipa = { version = "4.2", features = ["axum_extras", "actix_extras", "uuid", "chrono", "url"] }
 serde_json = "^ 1.0"
 itertools = "0.13"
 % if cargo.get('is_executable', False):
--- a/src/generator/templates/Cargo.toml.mako
+++ b/src/generator/templates/Cargo.toml.mako
@@ -62,5 +62,5 @@ version = "${util.crate_version()}"
 [features]
 yup-oauth2 = ["google-apis-common/yup-oauth2"]
 default = ["yup-oauth2"]
-utoipa = ["dep:utoipa", "utoipa/url"]
+utoipa = ["dep:utoipa"]
 % endif
--- a/src/generator/templates/api/api.rs.mako
+++ b/src/generator/templates/api/api.rs.mako
@@ -30,6 +30,7 @@ use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::time::sleep;
 use tower_service;
 use serde::{Serialize, Deserialize};
+use utoipa::ToSchema;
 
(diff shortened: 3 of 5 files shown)


that-ambuj:
For context, we are using this library and to locally derive `utoipa::ToSchema` on external types is very hacky and we have to maintain a copy of the external types(which are too large for this crate) and since the versions don't mean anything here, it is very hard to keep them in sync. The hack in question is mentioned here https://github.com/juhaku/utoipa/issues/507

src/generator/lib/util.py
```diff
@@ -1199,7 +1199,7 @@ def unique(
     original: str,
     desired: str,
     attempts: int = 0,
-    assigned: dict[str, str] = {},
+    assigned: Dict[str, str] = {},
```
Byron, line 1202: Could you educate me on this change?

src/generator/templates/Cargo.toml.mako
```diff
@@ -31,6 +31,7 @@ hyper-rustls = "0.25.0"
 ## Must match the one hyper uses, otherwise there are duplicate similarly named `Mime` structs
 mime = "^ 0.3.0"
 serde = { version = "^ 1.0", features = ["derive"] }
+utoipa = { version = "^4.2", features = ["axum_extras", "uuid", "chrono", "url"] }
```
Byron, line 34: There is indeed no way this would be added without being optional, for use with a feature toggle. Putting this in should be no problem though.

Byron (changes requested):
Thanks for bringing this up, and adding support for `utoipa` is certainly very useful. In a way, the API generator now goes full circle, from spec to code, back to (maybe some other) spec.

For this to be mergable, a feature toggle would have to be added though, just like you predicted.

Thanks for your consideration.

src/generator/lib/util.py
that-ambuj: Oh, as the README suggested, I used python 3.8.19 but it was not compiling because of the wrong type.
The newer versions use `dict` while the older versions use `Dict`, same goes with `List`. 

In order to do `make`, I had to make this change so it works are expected with python 3.8.

I might be wrong here but I added this for compatibility with the README and the python version. We can remove this but I don't know if other developers will face the same issue.

Can you confirm the python version on your machine and if the `make` works?

src/generator/templates/Cargo.toml.mako
that-ambuj: That's why it's a draft, I just did it hard and fast to see if it actually works for my use case.

src/generator/lib/util.py
Byron: I assume that older python interpreters are also fine with `Dict`, as otherwise this would be a breaking change in Python itself.
So let's keep what's most compatible.

that-ambuj:
Hey @Byron, could you tell me where do these `dependencies` come from in the below snippet?

https://github.com/Byron/google-apis-rs/blob/7b70b66c634ff7c798cda36fc68f5b99ae5a703a/src/generator/templates/Cargo.toml.mako#L41-L59

Byron:
They are probably from here:

https://github.com/Byron/google-apis-rs/blob/7b70b66c634ff7c798cda36fc68f5b99ae5a703a/etc/api/type-api.yaml#L28-L33

that-ambuj:
@Byron Do you think it would be a good idea to enable `utoipa/url` in the default feature set?

that-ambuj:
Would this be a good place to add documentation about this feature? or should it be mentioned separately as a `Crate Feature`?
https://github.com/Byron/google-apis-rs/blob/7b70b66c634ff7c798cda36fc68f5b99ae5a703a/src/generator/templates/api/lib/lib.mako#L69-L70

Byron:
> @Byron Do you think it would be a good idea to enable `utoipa/url` in the default feature set for `utoipa`?

I don't know - if in doubt don't enable it, as users of `utoipa` can always enable it in their own tree, along with any other features they might need.



> Would this be a good place to add documentation about this feature? or should it be mentioned separately as a `Crate Feature`?
> 
> https://github.com/Byron/google-apis-rs/blob/7b70b66c634ff7c798cda36fc68f5b99ae5a703a/src/generator/templates/api/lib/lib.mako#L69-L70

I'd probably put it at the very end of the library documentation, and cal it `Cargo Features`.

Besides that, the PR seems to be in a mergeable state, thanks a lot!

PS: It would be nice to see how these docs actually look like, in case you can post a screenshot for instance.

that-ambuj:
We only need to enable the features that are used by this library crate. I can make `utoipa/url` a separate feature but since we are using `url` crate in the library, I would argue that it should enabled in the default `utoipa` feature set. 

I'm really not sure where and what the `url` crate is used for. I'll just test it in my project if the crate works without that feature for utoipa.

that-ambuj:
Just confirmed, it works without the `url` feature. I should remove it from the `utoipa` feature set.

that-ambuj:
image

@Byron This is what the docs look like with `cargo doc`.

Byron:
Thanks! 

The PR looks like it can be merged now, thanks a lot for working on this.

To wrap it up, would you have a screenshot of the **utoipa** output? I have never seen it, and basically don't know what this is doing.

src/generator/lib/util.py
Byron: Could you respond to this?
Will `Dict` be backward compatible?

that-ambuj:
image

This is a screenshot from the swagger ui. It is generated from a big `openapi.json` file(with help of `utoipa`). The `openapi.json` can be used to generate client SDK using a tool like [OpenAPI generator](https://openapi-generator.tech). You can also see a sample of this at https://petstore.swagger.io .

src/generator/lib/util.py
that-ambuj: Sorry, I thought you wanted to keep this change. Yes, `Dict` is backwards compatible but `dict` is not. So, people using python 3.9+ should not have any issues as I have checked with version 3.8 and 3.12.

Byron: Great, thank you!

Byron:
So it does come full circle! There of course will be a good reason for doing that and not using the API definition that was used as source in the first place, but maybe it's subtly incompatible.

Maybe what would have helped to do the same is to keep a copy of the API definition JSON in the crate itself or make it very easy to download?

that-ambuj:
From my use case standpoint, using the swagger json instead of this feature would achieve the same result. utoipa generated the specs from structs and does not keep a copy of openapi.json. Even if it did, we would have to manually edit the file and put the required parts into the json file, which is very prone to human error and unreliable.

With this feature we can automatically generate the specs for only the types we need. 

Google uses the [JSON Schema](https://json-schema.org) for it's APIs, which I found out today. We could use codegen for that standard instead but this crate works more reliably than the codegen tools I've used, so I'm still very sceptical of them. 

P.S: You could check out https://github.com/Marwes/schemafy/ to see if it does anything better than your implementation. And I doubt if it's any better. And again, thanks for your good work and saving me and my teammates a lot of time.

that-ambuj:
Just checked, there are no open source or actively maintained tools to generate code from Google's Discovery format.

Byron:
Thanks for elaborating, that's all very interesting to me :)!

For a moment there I thought that Google doesn't provide bindings to their APIs in Rust because it's trivial to generate them using Swagger and Co., but from your experience that doesn't seem to be the case at all.
Maybe one day… .

PR #513, opened 2024-06-26
Update the API schemas and regenerate the crates

IvanUkhov:
Hello,

My actual intention is to ask to publish the Spanner crate. To this end, I have updated the schemas via

```shell
rm -f .api.deps .cli.deps && FETCH_APIS=1 make update-json -j8
```

and regenerated all crates (although had to add `bigtableadmin` to the black-list) via

```shell
make cargo-api ARGS=check
```

Please let me know if anything else is required to streamline the publishing. Thank you!


Byron:
Thanks a lot for the update and the instructions.

I will follow them myself (#514) and perform the API update on your behalf as most commits here aren't reviewable.

While doing so, I noticed that I got two 404 and ten 502, which is new to me. It also means that the Google API endpoint informs about APIs that can't be reached, and I am ignoring this.

In any case, the latest version of spanner should soon be available on crates.io as it feels like it's time for releasing all the improvements.

Byron:
Closing as superseded by #514.

Byron:
I changed my mind and only published spanner, but here it is: https://crates.io/crates/google-spanner1/5.0.5+20240618

IvanUkhov:
Oh, yes, I have also noticed 404 and 502. Since the script continued, I assumed it could be ignored.

Regarding the publishing, it would help to republish all actually. I do use Spanner but in combination with Pub/Sub, Storage, and Secret Manager, and since the dependencies have been updated, you end up with several versions of the same second-order dependency. For instance:

https://github.com/Byron/google-apis-rs/commit/46107ff49371fa97ba4fedd0bf09498f23a98337

But of course, you know best. Perhaps you are waiting for some other update before republishing. For me, it is non-blocking, as I can pull from by branch for the time being.

Thank you!

Byron:
That's fair enough, let's just go through with it then - I have been avoiding it but a mid-of-year publish probably isn't the worst thing to do.

IvanUkhov:
Thank you so much!

PR #515, opened 2024-07-24
Add aiplatform apis

stevemk14ebr:
Adds

* aiplatform v1
* aiplatform v1beta1

PR #518, opened 2024-08-14
Add cloudcommerceprocurement API

banool:
Adds:
- cloudcommerceprocurement v1

I tried to do the same thing as https://github.com/Byron/google-apis-rs/pull/515 did.

I was able to use this crate successfully in my own project (private repo unfortunately).

### DIFF
--- a/apis.json
+++ b/apis.json
@@ -4,6 +4,21 @@
   "items": [
     {
       "kind": "discovery#directoryItem",
+      "id": "cloudcommerceprocurement:v1",
+      "name": "cloudcommerceprocurement",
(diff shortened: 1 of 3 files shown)


Byron (changes requested):
Thanks a lot!

Could you remove the generated code though? I'd generate it myself later when publishing the crates.

Thanks again.

banool:
Done, I removed everything under `gen/`

Byron:
Thank you - here is the release: https://crates.io/crates/google-cloudcommerceprocurement1

PR #523, opened 2024-10-01
Make UserDraftCreateCall::doit public

kew96:
UserDraftCreateCall::doit is currently private which prevents performing the operation

### DIFF
--- a/gen/gmail1/src/api.rs
+++ b/gen/gmail1/src/api.rs
@@ -3384,7 +3384,7 @@ where
 
 
     /// Perform the operation you have build so far.
-    async fn doit<RS>(mut self, mut reader: RS, reader_mime_type: mime::Mime, protocol: client::UploadProtocol) -> client::Result<(hyper::Response<hyper::body::Body>, Draft)>
+    pub async fn doit<RS>(mut self, mut reader: RS, reader_mime_type: mime::Mime, protocol: client::UploadProtocol) -> client::Result<(hyper::Response<hyper::body::Body>, Draft)>
 		where RS: client::ReadSeek {
         use std::io::{Read, Seek};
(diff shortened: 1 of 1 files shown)


Byron (commented):
These are private as they have public counterparts:

```Rust
    /// Upload media in a resumable fashion.
    /// Even if the upload fails or is interrupted, it can be resumed for a
    /// certain amount of time as the server maintains state temporarily.
    /// 
    /// The delegate will be asked for an `upload_url()`, and if not provided, will be asked to store an upload URL
    /// that was provided by the server, using `store_upload_url(...)`. The upload will be done in chunks, the delegate
    /// may specify the `chunk_size()` and may cancel the operation before each chunk is uploaded, using
    /// `cancel_chunk_upload(...)`.
    ///
    /// * *multipart*: yes
    /// * *max size*: 36700160
    /// * *valid mime types*: 'message/*'
    pub async fn upload_resumable<RS>(self, resumeable_stream: RS, mime_type: mime::Mime) -> client::Result<(hyper::Response<hyper::body::Body>, Draft)>
                where RS: client::ReadSeek {
        self.doit(resumeable_stream, mime_type, client::UploadProtocol::Resumable).await
    }
    /// Upload media all at once.
    /// If the upload fails for whichever reason, all progress is lost.
    ///
    /// * *multipart*: yes
    /// * *max size*: 36700160
    /// * *valid mime types*: 'message/*'
    pub async fn upload<RS>(self, stream: RS, mime_type: mime::Mime) -> client::Result<(hyper::Response<hyper::body::Body>, Draft)>
                where RS: client::ReadSeek {
        self.doit(stream, mime_type, client::UploadProtocol::Simple).await
    }
```

Closing, as it makes no sense to edit a generated file - it will be overwritten next time the generator runs.

PR #524, opened 2024-10-02
Update to Hyper 1

### LINKED ISSUE #459: Upgrade to Hyper 1.0
See https://seanmonstar.com/blog/hyper-v1/. The API changed, so this might mean a new major version in semver for google-apis-rs as well.

IvanUkhov:
Closes #459.


IvanUkhov:
@Byron, just wanted to get your early feedback. I have managed to make the API crates compile. At least, it passes for:

```shell
make storage1-clean storage1 storage1-cargo ARGS=test
```

It remains to update the CLI template.

Is this something that would would be willing to accept, or is it not the direction you want to take?

Byron:
Thank you, the work looks very promising and I'd be happy if we could see it through.

My recommendation is to get CI green first - the issues there seem to be unrelated to the CLI for now.

And regarding the CLI, I believe to remember that they didn't compile before your change either, it degenerated (maybe CI isn't actually testing these), and getting them to work again would be greatly appreciated - maybe there are still useful to some.

IvanUkhov:
Now the CI fails at the CLIs. I will then tackle them next.

IvanUkhov:
I added storage1 to the CI list to show the error I am facing (was passing for the other ones on the list). Please take a look. Looks like there is some confusion regarding `type` and `format`. An argument can be of type `string` and format `int64`, and the function is generated to take an integer while the CLI tries to pass a string. Do you have any ideas regarding what might be going on and how to address that?

src/generator/templates/api/lib/mbuild.mako
```diff
@@ -703,17 +698,6 @@ else {
             request_value_reader.seek(io::SeekFrom::Start(0)).unwrap();
             % endif
             let mut req_result = {
-            % if resumable_media_param:
```
Byron, line 706: Would this not be missing somewhere?

IvanUkhov: I was planning to ask you about this. It was getting in the way, but I could not find how it was actually used. It looks like dead code. The way I see it is that `should_ask_dlg_for_url ` is always false. Or what is going on here?

IvanUkhov: I was planning to ask you about this. The way I see it `should_ask_dlg_for_url` is always false. So that looked like a piece of dead code. Or what is going on here?

Byron: Right! I checked and the variable was false from the very beginning, so it could never have worked anyway.

Also something I just noticed is the 'assignment as side-effect of if-condition' dark magic that is performed here, and it's incredible that I was the author of this 😅.

In any case, I agree that under the circumstances, it should definitely be removed.

Byron (commented):
Thanks a lot for following up!

From my point of view this would be ready to merge, unless of course you'd like to perform more testing or try to figure out how to fix that last remaining CLI generation issue.

It's entirely up to you.

IvanUkhov:
I have plugged in a few Python linters and cleaned up trailing whitespace and such. Also, clarified the CI workflow.

I will try to regenerate all APIs and run the checks with the tests now.

IvanUkhov:
OK, I am able to generate all APIs and the following succeeds:

```shell
 make cargo-api ARGS=check
```

etc/api/type-api.yaml
```diff
@@ -25,9 +25,3 @@ make:
 cargo:
   keywords: [protocol, web, api]
   doc_base_url: https://docs.rs
-  dependencies:
```
Byron, line 28: It's a great idea to bring it all into the generator file.

src/generator/templates/deps.mako
```diff
@@ -110,6 +110,7 @@ ${api_common}: ${gen_root_stamp}
 ${gen_root_stamp}: $(MAKO_RENDER) ${' '.join(i[0] for i in sds)} ${api_json_inputs} $(MAKO_STANDARD_DEPENDENCIES)
 	@echo Generating ${api_target}
 	$(MAKO) -io ${' '.join("%s=%s" % (s, d) for s, d in sds)} ${post_processor_arg} --data-files ${api_json_inputs}
+	${'echo ' if gen_root.endswith('-cli') else ''}cargo fmt --manifest-path ${gen_root}/Cargo.toml
```
Byron, line 113:
```suggestion
	${'echo TODO make CLIs compile ' if gen_root.endswith('-cli') else ''}cargo fmt --manifest-path ${gen_root}/Cargo.toml
```

Byron (approved):
Thanks a lot for the exhaustive update and modernisation!

It's great that CI is now more granular, too. Maybe it could be split between API and CLI as well so it's more clear that only CI currently fails, and nothing else.

Maybe with all the time spent in the code-base, it will soon become more obvious why some CLIs don't yet build. It seemed like it must be a small thing… .

PR #526, opened 2024-10-07
Format the CLIs

IvanUkhov:
It was originally failing, but then I cleaned up the templates, and it started to pass. Forgot to remove the echo.


Byron (approved):
That's awesome, thanks so much!

Byron:
PS: Maybe it would be easy to you to split the CI further into API and CLI, so it's clear that currently only the CLI fails.

PR #527, opened 2024-10-10
Avoid processing incoming bodies unnecessarily

IvanUkhov:
Previously, I defined the body to be `Full<Bytes>`, which meant that the response would have to be fully received inside the API functions. However, that seems wasteful if one wants to do some streaming processing of the incoming data. This PR switches over to `http_body_util::combinators::BoxBody`, which is what is [recommended](https://hyper.rs/guides/1/upgrading/) in cases of having to accommodate different response types.

### DIFF
--- a/google-apis-common/Cargo.toml
+++ b/google-apis-common/Cargo.toml
@@ -19,7 +19,7 @@ chrono = { version = "0.4", default-features = false, features = ["clock", "serd
 http = "1"
 http-body-util = "0.1"
 hyper = { version = "1", features = ["client", "http2"] }
-hyper-util = "0.1"
+hyper-util = { version = "0.1", features = ["client-legacy", "http2"] }
 itertools = "0.13"
 mime = "0.3"
 percent-encoding = "2"
--- a/src/generator/templates/Cargo.toml.mako
+++ b/src/generator/templates/Cargo.toml.mako
@@ -32,7 +32,7 @@ clap = "2"
 http-body-util = "0.1"
 % endif
 hyper = "1"
-hyper-rustls = "0.27"
+hyper-rustls = { version = "0.27", default-features = false }
 hyper-util = "0.1"
 mime = "0.3"
 serde = { version = "1", features = ["derive"] }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,15 +7,22 @@ on:
     branches: [main]
 
 jobs:
-  meta-check:
-    name: Check the generative code
+  common-check:
+    name: Check the common code
     runs-on: ubuntu-latest
     steps:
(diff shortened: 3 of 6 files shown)


src/generator/templates/Cargo.toml.mako
```diff
@@ -32,7 +32,7 @@ clap = "2"
 http-body-util = "0.1"
 % endif
 hyper = "1"
-hyper-rustls = "0.27"
+hyper-rustls = { version = "0.27", default-features = false }
```
IvanUkhov, line 35: This is something extra. `hyper-rustls` has several default choices, which some might not like.

Byron: It looks like now [the legacy support](https://github.com/Byron/google-apis-rs/actions/runs/11277500609/job/31363793254?pr=527#step:4:587) is also gone maybe?

Byron (changes requested):
Thanks a lot for your help with this, it's much appreciated!

Now that it seems that CI caught a legitimate issue that is way too easily ignored, let's merge this PR only after the CLI related tests are in their own jobs so they can be expected to fail officially (I think CI can also be told so) so we are back at 'green CI by default'.

That should make further possibly even sweeping changes possible without adding unintended breakage.

IvanUkhov:
Oh, was not paying attention to the CLIs. I will take a look.

IvanUkhov:
Yeah, I think that one is a legitimate failure that should be eliminated. If I had noticed, I would not have let it in.

IvanUkhov:
I actually misread the error. It was not about the CLIs; it was about the APIs with the default features disabled. It should be resolved now.

google-apis-common/Cargo.toml
```diff
@@ -19,7 +19,7 @@ chrono = { version = "0.4", default-features = false, features = ["clock", "serd
 http = "1"
 http-body-util = "0.1"
 hyper = { version = "1", features = ["client", "http2"] }
-hyper-util = "0.1"
+hyper-util = { version = "0.1", features = ["client-legacy", "http2"] }
```
IvanUkhov, line 22: Here I do as on the line above. I do not think it matters if you activate `http1` or `http2`.

Byron:
Thanks a lot, it now looks like a 'good failure'. 

Do I understand correctly that you are not interested in splitting the CI jobs between API and CLI to prevent such issues in future?

IvanUkhov:
Sure! Let me do that.

Byron (approved):
Thanks again, particularly about the improvements to CI!
Now I am also curious if 'expects to fail' is something that CI can do, but then again, I still have hopes that the CLIs can be made to work or that those that aren't working can be excluded one day.

PR #510, opened 2024-06-11
Bump yup-oauth2 to v10

davidkuridza:
Version 10 of [yup-oauth2](https://github.com/dermesser/yup-oauth2) was released recently, this PR bumps the version here as well.

PR #529, opened 2024-10-15
chore: add places api v1

b0xtch:
Add support for places api

### DIFF
--- a/apis.json
+++ b/apis.json
@@ -4029,6 +4029,21 @@
     },
     {
       "kind": "discovery#directoryItem",
+      "id": "places:v1",
+      "name": "places",
+      "version": "v1",
+      "title": "Places API (New)",
+      "description": "",
+      "discoveryRestUrl": "https://places.googleapis.com/$discovery/rest?version=v1",
+      "icons": {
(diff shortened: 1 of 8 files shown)


Byron:
Thanks a lot!

b0xtch:
> Thanks a lot!

ofc qq on the publishing I am not sure what the procedure is! should I publish I don't see it in the CI either

Byron:
No, please don't publish it.

It doesn't show in CI as CI only tests a small subset of APIs and CLIs - otherwise it would never finish.

I will re-publish the crates in the coming days (it was [also requested elsewhere](https://github.com/Byron/google-apis-rs/pull/510#issuecomment-2412977165)).

b0xtch:
> No, please don't publish it.
> 
> It doesn't show in CI as CI only tests a small subset of APIs and CLIs - otherwise it would never finish.
> 
> I will re-publish the crates in the coming days (it was [also requested elsewhere](https://github.com/Byron/google-apis-rs/pull/510#issuecomment-2412977165)).

you got it, thanks 🙏

PR #535, opened 2025-01-31
Upgrade yup-oauth2 to v12

paolobarbolini:


Byron:
Changing generated files is of no use as these changes are overwritten next time they are regenerated.
In order to perform such an upgrade, please only change the generator code.

Thanks.

PR #537, opened 2025-02-01
Drop dependency on itertools

paolobarbolini:


Byron (approved):
Thanks a lot, looks good to me!
Unfortunately I could only estimate that it will work as there isn't a test for that, but by all means, the new version *should* produce the same result.

google-apis-common/src/lib.rs
```diff
@@ -414,22 +413,26 @@ impl Read for MultiPartReader<'_> {
...
                 let (headers, reader) = self.raw_parts.remove(0);
+
+                let mut encoded_headers = String::new();
+                for (k, v) in &headers {
+                    if !encoded_headers.is_empty() {
+                        encoded_headers.push_str(LINE_ENDING);
+                    }
+
+                    write!(encoded_headers, "{}: {}", k, v.to_str().unwrap())
+                        .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
```
paolobarbolini, line 426: Since Rust 1.74 this can be simplified to `io::Error::other`
https://doc.rust-lang.org/nightly/std/io/struct.Error.html#method.other

```suggestion
                        .map_err(|err| std::io::Error::other(err))?;
```

Byron: A good call, thank you!

PR #539, opened 2025-02-02
Reduce yup-oauth2 features

paolobarbolini:
Let's see what CI thinks of this...


Byron:
I think this PR needs a little more context so I can understand its progress, hence I have put it into draft mode. Please let me know when you think it's ready.

paolobarbolini:
This is now ready to be merged.

src/generator/templates/Cargo.toml.mako
```diff
@@ -32,7 +32,7 @@ clap = "2"
 http-body-util = "0.1"
 % endif
 hyper = "1"
-hyper-rustls = { version = "0.27", default-features = false }
+hyper-rustls = { version = "0.27", default-features = false, features = ["http2", "rustls-native-certs", "native-tokio"] }
```
Byron, line 35: How come these are enabled by default now?

Byron (changes requested):
Thank you. 

I think I pieced together what the PR is supposed to do, but would hope you could add comments (in [`document-features` style](https://docs.rs/document-features/latest/document_features/)) so it becomes clearer how these new features are supposed to be used, and to verify I actually do get what's intended.

Also, would you think the major version of the crates needs to be bumped? Isn't this a breaking change without adding the previous default-features back in (if that's even possible)?

src/generator/templates/Cargo.toml.mako
paolobarbolini: They were already enabled by `yup-oauth2` before, but since we now compile it without the `hyper-rustls` feature we have to enable it ourselves.

paolobarbolini:
> I think I pieced together what the PR is supposed to do, but would hope you could add comments (in [`document-features` style](https://docs.rs/document-features/latest/document_features/)) so it becomes clearer how these new features are supposed to be used, and to verify I actually do get what's intended.

Done

> Also, would you think the major version of the crates needs to be bumped

They would have to anyways because upgrading `yup-oauth2` in #536 is already a breaking change.

Byron (changes requested):
Thanks a lot!

Thanks for clarifying the breaking change as well.

Then I think we can merge once the `cargo.build_version` was updated in `shared.yml`.

Byron (approved):
Thanks a lot!

paolobarbolini:
Is there anything left to be done for this to be merged?

Byron:
Apologies, I somehow failed to hit the merge button.

Byron:
At this point I think I'd have to do that myself. Something that should be done before is to update all APIs and maybe provide new crates in the process (for new APIs).
I'd like to wait more to let it settle though - users of these crates are probably quite used to using vendored versions anyway given how infrequently they update.
Maybe as an intermediate step, documentation could be added to show how to rebuild an API crate and use it via vendoring?

paolobarbolini:
Given how easy it is to make new releases in the Rust world and how easy it is for users who don't want to upgrade immediately to stand back, I'd rather not vendor. I think Rust makes it easy to have an upstream-first approach but that stops working when new releases then take too long to come out.

PR #540, opened 2025-02-21
Un-blacklist google-bigquery2

main--:
While outright removing the recursive fields is not ideal, this is still a lot better than just leaving the google-bigquery2 crate with no updates for the foreseeable future.


Byron (approved):
Thanks a lot, I am impressed this works at it seems a bit magical to me.
But… most of the python code seems magical to me by now 😅.

PR #554, opened 2025-06-29
fix(gmail1): Correct OAuth2 scope precedence for message operations

mazhewitt:
## Problem

Users experience "Missing access token for authorization" errors when calling individual Gmail message operations with valid OAuth2 tokens that have standard Gmail scopes:

- ✅ `messages().list()` works correctly
- ❌ `messages().get()` fails with 403 authorization error
- ❌ `threads().get()` fails with 403 authorization error  
- ❌ `messages().attachments().get()` fails with 403 authorization error

This occurs even when users have proper `gmail.readonly`, `gmail.modify`, or `gmail.compose` scopes.

## Root Cause

**Inconsistent OAuth2 scope defaults** in the Gmail API discovery document:

- `messages.list()` correctly defaults to `gmail.readonly` scope
- `messages.get()`, `threads.get()`, `attachments.get()` incorrectly default to `gmail.addons.current.message.readonly` scope

The addon-specific scope is highly restrictive and intended only for Gmail add-ons, not general API usage. The Rust code generator picks the first applicable scope from the discovery document as the default, and the addon scopes were listed before the standard scopes.

## Solution

**1. Fixed Gmail API Discovery Document**
- Reordered scopes in `/etc/api/gmail/v1/gmail-api.json` to prioritize standard Gmail scopes
- Moved `gmail.readonly` before addon-specific scopes in three methods:
  - `users.messages.get` (lines 936-943)
  - `users.threads.get` (lines 2905-2912)  
  - `users.messages.attachments.get` (lines 1339-1346)

**2. Regenerated Gmail API Code**
- Used `make gmail1-clean && make gmail1` to regenerate clean API code
- All three methods now default to `Scope::Readonly` instead of `Scope::AddonCurrentMessageReadonly`

## Verification

- ✅ Gmail API compiles successfully
- ✅ All methods now consistently use `gmail.readonly` as default scope
- ✅ Generated code shows proper scope usage in `gen/gmail1/src/api.rs`
- ✅ Backward compatibility maintained (all scopes still available)

## Impact

**Fixes**:
- Individual message fetching now works with standard OAuth2 tokens
- Thread fetching now works with standard OAuth2 tokens
- Attachment fetching now works with standard OAuth2 tokens

**Maintains**:
- All existing functionality and API compatibility
- User ability to override scopes via `.add_scope()` if needed
- Proper OAuth2 scope hierarchy and security model

## Files Changed

- `etc/api/gmail/v1/gmail-api.json` - Discovery document scope reordering (source fix)
- `gen/gmail1/src/api.rs` - Regenerated API with corrected default scopes
- `gen/gmail1/Cargo.toml`, `gen/gmail1/README.md`, `gen/gmail1/src/lib.rs` - Regenerated metadata

## Testing

Users can verify the fix by:
1. Authenticating with standard Gmail scopes (`gmail.readonly`, `gmail.modify`, etc.)
2. Successfully calling `messages().get()`, `threads().get()`, and `attachments().get()`
3. Confirming no "Missing access token for authorization" errors

This addresses the root cause ensuring the fix persists through future code generation cycles.

### DIFF
--- a/etc/api/gmail/v1/gmail-api.json
+++ b/etc/api/gmail/v1/gmail-api.json
@@ -936,12 +936,12 @@
                             },
                             "scopes": [
                                 "https://mail.google.com/",
-                                "https://www.googleapis.com/auth/gmail.addons.current.message.action",
-                                "https://www.googleapis.com/auth/gmail.addons.current.message.metadata",
-                                "https://www.googleapis.com/auth/gmail.addons.current.message.readonly",
+                                "https://www.googleapis.com/auth/gmail.readonly",
                                 "https://www.googleapis.com/auth/gmail.metadata",
                                 "https://www.googleapis.com/auth/gmail.modify",
-                                "https://www.googleapis.com/auth/gmail.readonly"
+                                "https://www.googleapis.com/auth/gmail.addons.current.message.action",
+                                "https://www.googleapis.com/auth/gmail.addons.current.message.metadata",
+                                "https://www.googleapis.com/auth/gmail.addons.current.message.readonly"
                             ]
                         },
                         "import": {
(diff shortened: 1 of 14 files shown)


Byron:
Wow, thanks a lot for this! I haven't seen CI green here for a while and didn't think it was possible.

I think this PR can be merged as is, particularly because it fixes CI, however, when doing so I can promise you that all changes in `etc/` and `gen/` will be overwritten the next time the APIs are fetched from the Google server.
There isn't a mechanism to configure the scopes or rearrange them yet, but there is a way to set overrides, merging changes  from one yaml file to the parsed Google API description.

Byron (approved):

PR #556, opened 2025-07-23
fix(homegraph1): Un-blacklist by adding 'async' to reserved words

zopieux:
I'm not sure why the solution to `async` being a keyword was blacklisting instead of adding `async` to the long list of reserved keywords. So this PR removes homegraph1 from the blacklist and adds the gen back.

I would welcome guidance on what to do to "release" a new version of the crate, in this PR or as a follow-up.

Also: fixes the comment about the blacklist in `shared.yaml`, which somehow moved to a random item within.


Byron:
Thanks a lot for sorting this out! It looks like I forgot how to operate this machine and failed to what was done in this PR.

Thanks also for cleaning up the comment, great catch.

> I would welcome guidance on what to do to "release" a new version of the crate, in this PR or as a follow-up.

Thanks for refraining from publishing it yourself - I published it now.

PR #570, opened 2026-02-08
feat(generator): support impl Into<String> in setters

raushan728:
*   Updated templates to use `impl Into<String>` for string-based setters.
*   Improves DX by allowing string literals without manual `.to_string()`.
*   Verified with `make youtube3` and `cargo check`.


Byron (approved):
Looks good to me, thanks!

PR #573, opened 2026-04-06
Fix RUSTSEC-2025-0134: update yup-oauth2 dependency from v11 to v12 i…

mansiverma897993:
Fixes RUSTSEC-2025-0134 by updating the `yup-oauth2` dependency to a newer version, which removes the usage of the unmaintained `rustls-pemfile` crate.

Verified that `cargo audit` no longer reports the advisory.


Byron:
Bad bot: edit generator files, not files that are generated and explicitly state that they shall not be edited.

PR #574, opened 2026-06-01
Fix formatting links in README

c-git:
I wasn't sure if that was the intended display of that oauth info but was the only one that made sense to me. Mostly was trying to fix the display of the other links


Byron:
Thanks for the hint!

I fixed this on main directly - strange this happened in the first place.

PR #575, opened 2026-06-15
docs: fix generated rustdoc links

### LINKED ISSUE #297: Sheets4 Api docs link broken
Hey Sebastian

Thanks a lot for your work on this project!

If I click on any link on the Readme of https://github.com/Byron/google-apis-rs/tree/main/gen/sheets4

It gives a 404 on the Rust docs website. Would be happy to make a PR to fix it tho!

wyf027:
Closes #297.

## Summary
- update generated README Rustdoc links from `api::Type` to current docs.rs paths like `api/struct.Type.html`
- link common marker traits to `google-apis-common/latest` documentation
- refresh `gen/sheets4/README.md` so the reported Sheets API links resolve now

## Verification
- `git diff --check`
- `PYTHONPATH=src uv run --with mako --with inflect python - <<'PY' ... Template(filename='src/generator/templates/api/lib/lib.mako') ... PY`
- checked 28 unique docs.rs links in `gen/sheets4/README.md`; all updated google-sheets4/google-apis-common links returned HTTP 200

Note: an existing `yup-oauth2/*/trait.AuthenticatorDelegate.html` link still 404s because current yup-oauth2 no longer exposes that trait; this PR stays focused on the generated Rustdoc links reported for the Sheets API README.

### DIFF
--- a/gen/sheets4/README.md
+++ b/gen/sheets4/README.md
@@ -11,10 +11,10 @@ Everything else about the *Sheets* *v4* API can be found at the
 [official documentation site](https://developers.google.com/workspace/sheets/).
 # Features
 
-Handle the following *Resources* with ease from the central [hub](https://docs.rs/google-sheets4/7.0.0+20251215/google_sheets4/Sheets) ...
+Handle the following *Resources* with ease from the central [hub](https://docs.rs/google-sheets4/7.0.0+20251215/google_sheets4/api/struct.Sheets.html) ...
 
-* [spreadsheets](https://docs.rs/google-sheets4/7.0.0+20251215/google_sheets4/api::Spreadsheet)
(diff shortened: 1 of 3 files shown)


Byron (changes requested):
Thanks a lot.

I validated some of these newly generated links and they work, whereas the old ones don't.
Why it works a little bit puzzling to me still, and I'd like to see `lib.rs` regenerated as well as `${lib.docs(...)` is called from there as well.

wyf027:
Addressed the requested `lib.rs` regeneration in d3623ee29.

What changed:
- Regenerated `gen/sheets4/src/lib.rs` so the hub references now use `api::Sheets`, matching the generator/template behavior already reflected in `gen/sheets4/README.md`.
- Left unrelated regenerated `api.rs` formatting noise out of the commit.

Validation:
- `cargo check --manifest-path gen/sheets4/Cargo.toml` passed.
- `cargo doc --manifest-path gen/sheets4/Cargo.toml --no-deps` passed; it still emits pre-existing generated API warnings for unrelated Sheets schema text links/bare URL.

Byron (approved):
Thanks a lot!

I still don't really understand how this works. To my mind that change should duplicate a line but apparently it doesn't. I also don't want to spend more time on it (and validate it locally), so let's just merge - in the worst case links in markdown files are more broken or docs don't build, something that CI might actually be doing as well.

PR #576, opened 2026-07-02
Add google-health4 crate for Google Health API v4

null-sleep:
Generates a Rust client for https://developers.google.com/health (health:v4), which is not yet packaged by this project.

Steps taken:
1. Registered the API in etc/api/shared.yaml under `manually_added`, pointing at its discovery doc: https://health.googleapis.com/$discovery/rest?version=v4
2. Fetched that discovery doc to etc/api/health/v4/health-api.json and normalized it with etc/bin/sort_json_file.py.
3. Regenerated etc/api/api-list.yaml via etc/bin/api_version_to_yaml.py so the build knows about `health: [v4]`.
4. Ran `make health4` to render the crate from the discovery doc (builds the preproc helper and Python venv on first run).
5. Verified with `make health4-cargo ARGS=check` (clean, no blacklist entries or overrides needed).

To consume the crate from another Rust project, add a path or git dependency since `gen/` is excluded from this workspace and google-health4 depends on google-apis-common via a relative path:

    [dependencies]
    google-health4 = { git = "https://github.com/null-sleep/google-apis-rs", package = "google-health4" }
    google-health4 = { git = "https://github.com/null-sleep/google-apis-rs", branch = "add-health-v4", package = "google-health4" } \\ OR
    google-health4 = { path = "/path/to/google-apis-rs/gen/health4" } \\ OR

### DIFF
--- a/etc/api/api-list.yaml
+++ b/etc/api/api-list.yaml
@@ -332,6 +332,8 @@ api:
     - v1
     groupssettings:
     - v1
+    health:
+    - v4
     healthcare:
     - v1
     - v1beta1
--- a/etc/api/health/v4/health-api.json
+++ b/etc/api/health/v4/health-api.json
@@ -1,0 +1,6002 @@
+{
+    "auth": {
+        "oauth2": {
+            "scopes": {
+                "https://www.googleapis.com/auth/cloud-platform": {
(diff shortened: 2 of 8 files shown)


null-sleep:
Oops, supposed to be for my fork 😅

Byron:
This reminds me that at some point I should probably refetch all APIs and include the new ones if possible.

